PR #1, opened 2020-11-02
Initial commit for odh-deployer

tmckayus:
The odh-deployer is moving from a subdirectory
in odh-operator-allinone to its own repository

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+
+ENV HOME /root
+
+RUN microdnf update -y && \
+    microdnf install -y \
+      bash \
+      tar \
+      gzip \
+    && microdnf clean all && \
+    rm -rf /var/cache/yum
+
+ADD https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz $HOME/ 
+RUN tar -C /usr/local/bin -xvf $HOME/oc.tar.gz && \
+    chmod +x /usr/local/bin/oc
+
+COPY deploy.sh /
+COPY opendatahub.yaml /
+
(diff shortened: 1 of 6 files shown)


tmckayus:
@crobby ptal

tmckayus:
just may want to drop appropriate pipeline config in here before merging

harshad16 (commented):
- we need to inclued the `.aicoe-ci.yaml` file to pick up the details regarding the builds.
```
check:
  - thoth-build
build:
  build-stratergy: "Dockerfile"
  registry: "quay.io"
  registry-org: "modh"
  registry-project: "odh-deployer"
  registry-secret: "modh-pusher-secret"
```

Makefile
```diff
@@ -0,0 +1,21 @@
+CONTAINER_REPO := quay.io/opendatahub
```
harshad16, line 1: if we are going to use the makefile, Shouldn't this be pointing to the `modh` quay.

harshad16 (commented):
lgtm,
just a minor comment left for the makefile.

Makefile
tmckayus: probably, I'll take a look.  I copied the contents from odh-operator-allinone and didn't review, thanks for looking!

Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+
+ENV HOME /root
```
vpavlin, line 3: Does this assume the container will run as root? (and the use of `/` later in general?)

Can we make it more "OpenShift SCC" friendly?

tmckayus: let's ping @crobby  and take up in another PR. This was a migration of existing sources from odh-operator-allinone

PR #3, opened 2020-11-03
Add --namespace flag to oc apply when necessary

tmckayus:
When deploying an object into a namespace that differs from
a namespace value in the object, a --namespace flag specifying
the value in the object is necessary to avoid an error.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -16,12 +16,21 @@
 
 set -e -o pipefail
 
-ODH_PROJECT=${ODH_CR_NAMESPACE:-"opendatahub"}
+# this is the value hardcoded in the CR yaml
+CR_NAMESPACE_VALUE=opendatahub
+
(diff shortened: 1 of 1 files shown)


tmckayus:
The error looks like this:

error: the namespace from the provided object "opendatahub" does not match the namespace "test-odh". You must pass '--namespace=opendatahub' to perform this operation.

You can easily test this locally with an oc login using the deploy.sh from the commandline

1) make deploy.sh executable
2) edit deploy.sh and set the CR path to "opendatahub.yaml" (or copy opendatahub.yaml to /)
3) create a new namespace like test-odh
4) export ODH_CR_NAMESPACE=test-odh
5) run deploy.sh
6) unset ODH_CR_NAMESPACE and run again

(if you really want to be sure, take out the namespace arg from deploy.sh and verify the error when deploying in test-odh :) )

tmckayus:
@crobby @vpavlin ptal

tmckayus:
hmm, hold on, something seems to be not working ...

PR #11, opened 2021-02-04
Changes to specify user 1001 and use a dir other than /root for HOME

### LINKED ISSUE #2: Review docker file for root usage
The Dockerfile sets the HOME env var to /root, does this imply the image is running as root? Does it need to?  If it isn't running as root, HOME should be set somewhere else.

crobby:
Closes #2


tmckayus:
lgtm

PR #20, opened 2021-02-14
Add oauth proxy to prometheus, scale down grafana and update namespace

maulikjs:


grafana.yaml
```diff
@@ -111,7 +111,7 @@ metadata:
   name: grafana
   namespace: redhat-monitoring
 spec:
-  replicas: 3
+  replicas: 1
```
anishasthana, line 114: any reason we're scaling it down? I thought HA is always desirable.

prometheus.yaml
```diff
@@ -18,21 +18,59 @@ spec:
...
+          "name": "redhat-monitoring", "namespace": "redhat-monitoring"}}'
+        - -tls-cert=/etc/tls/private/tls.crt
+        - -tls-key=/etc/tls/private/tls.key
+        - -client-secret-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+        - -cookie-secret-file=/etc/proxy/secrets/session_secret
+        - -openshift-ca=/etc/pki/tls/cert.pem
+        - -openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+        - -client-id=system:serviceaccount:redhat-monitoring:prometheus
+        - -skip-auth-regex=^/metrics
+        image: quay.io/openshift/origin-oauth-proxy:4.6.0
```
anishasthana, line 40: Could we just use the sme image as what is used for Grafana?

prometheus.yaml
```diff
@@ -59,6 +97,24 @@ spec:
...
+          defaultMode: 420
+          secretName: prometheus-tls
+      - name: prometheus-proxy
+        secret:
+          defaultMode: 420
+          secretName: prometheus-proxy
+---
+apiVersion: v1
+data:
+  session_secret: "uTgRq5qg4GlwsIXrAePXjTqz5DBhKqhERfGVOmyS"
```
anishasthana, line 111: We'll need to generate this "live" eventually. I'm writing some bash to hopefully make it easier.

prometheus.yaml
```diff
@@ -167,3 +235,16 @@ roleRef:
   kind: ClusterRole
   name: cluster-reader
   apiGroup: rbac.authorization.k8s.io
+# ---
```
anishasthana, line 238: Just leave this commented bit out for now.

anishasthana (changes requested):

grafana.yaml
maulikjs: Well, this is more of a load balancing thing rather than an HA thing at least that's what it was for telemetry given the number of users fetching dashboards. I assume we arent expecting that heavy a workload also since they will all be going to the same Prometheus it doesn't really help.

prometheus.yaml
maulikjs: 👍 yup yup

maulikjs: I think we already have the bash scripts for it in our dh-monitoring repo.

prometheus.yaml
maulikjs: yeah, I didnt realize there was a newer version out there.

grafana.yaml
anishasthana: Well having at least one replica is important in case the node grafana is running on has some issues. We can lower it to two if you really want.

prometheus.yaml
```diff
@@ -18,21 +18,59 @@ spec:
     spec:
       serviceAccountName: prometheus
       containers:
+      - args:
+        - -provider=openshift
+        - -https-address=:9091
+        - -http-address=
```
anishasthana, line 24: if the entry is empty should we just remove this?

anishasthana (commented):
We can merge the changes in once my last comment is resolved/rejected.

anishasthana (approved):

crobby:
Can someone also fire-up a cherry-pick of this to the v1.0.0 branch?  Thanks

PR #24, opened 2021-02-15
Update ODH version, remove bundled monitoring stack and enable oauth for prometheus

maulikjs:


anishasthana:
I think something went :boom: with the cherry pick

PR #26, opened 2021-02-16
Move Monitoring artifacts to monitoring folder

anishasthana:


Dockerfile
```diff
@@ -14,33 +14,32 @@ RUN microdnf update -y && \
...
       org.label-schema.name="ODH deployer" \
-      org.label-schema.vcs-ref="${vcs}" \
+      org.label-schema.vcs-ref="$vcs" \
       org.label-schema.vendor="Red Hat" \
-      org.label-schema.version="${version}"
+      org.label-schema.version="$version"
 
-ENTRYPOINT [ "/deploy.sh" ]
+WORKDIR $HOME
+ENTRYPOINT [ "/home/deployer/deploy.sh" ]
```
anishasthana, line 43: Thought here -- is it a bad idea to add $HOME to the $PATH variable?

crobby (changes requested):
The entrypoint is still set to /deploy.sh  I think that will cause problems.

anishasthana:
The entrypoint is set to ./deploy.sh . 
image

crobby (approved):
lgtm.  My problem was.....my problem.

crobby:
> The entrypoint is still set to /deploy.sh I think that will cause problems.

my mistake.....this is good

crobby (approved):
Seems to be working well, lgtm.

PR #28, opened 2021-02-17
Add alertmanager and proxy

maulikjs:


monitoring/prometheus.yaml
```diff
@@ -114,36 +188,134 @@ metadata:
...
 type: Opaque
 
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: alertmanager-proxy
+  namespace: redhat-odh-monitoring
+data:
+  session_secret: "Mqgtyv963kLX7NlVopzM3NnjonxKYBuyN6VbLVH6"
```
anishasthana, line 198: Can you generate this on the fly as well? You can look at deploy.sh to see how I did it.

maulikjs: ooh I didnt realize that was good to go, on it

monitoring/prometheus.yaml
```diff
@@ -100,50 +166,146 @@ spec:
...
         secret:
           defaultMode: 420
           secretName: prometheus-proxy
+      - name: alertmanager-proxy
+        secret:
+          defaultMode: 420
+          secretName: alertmanager-proxy
 ---
 apiVersion: v1
-data:
```
anishasthana, line 109: This needs to be deleted right?

anishasthana (approved):

PR #29, opened 2021-02-18
Move monitoring to a separate folder, add alertmanager and proxy, auto-generate session-secrets

maulikjs:


crobby (approved):
works for me

PR #34, opened 2021-02-19
Allow prometheus to scrape metrics from other namespaces and federate centralized k8s prometheus metrics

maulikjs:


monitoring/prometheus.yaml
```diff
@@ -212,7 +212,26 @@ data:
...
+      honor_labels: true
+      metrics_path: '/federate'
+
+      params:
+        'match[]':
+          - '{job=~"[A-z].*"}'
+
+      static_configs:
+        - targets:
+          - "prometheus-k8s-openshift-monitoring.apps.rhods-idh.dev.datahub.redhat.com"
```
anishasthana, line 233: This endpoint is hardcoded

monitoring/prometheus.yaml
```diff
@@ -377,16 +396,46 @@ spec:
     kind: Service
     name: prometheus
   wildcardPolicy: None
+# ---
```
anishasthana, line 399: can we remove this?

PR #38, opened 2021-02-24
Update prometheus flag, update prometheus SA permissions

maulikjs:


anishasthana:
The bash seems unchanged here?

PR #63, opened 2021-03-19
Use the Authentication overlay for odh-dashboard

accorvin:
This points our kfdef at the Authentication overlay for the ODH
Dashboard, which will configure it with an OAuth proxy to authenticate
and authorize users.

### DIFF
--- a/opendatahub.yaml
+++ b/opendatahub.yaml
@@ -33,6 +33,8 @@ spec:
         path: jupyterhub/notebook-images
     name: notebook-images
   - kustomizeConfig:
+      overlays:
+        - authentication
       repoRef:
         name: manifests
         path: odh-dashboard


LaVLaS:
LGTM. This was the same overlay I enabled when testing dashboard authentication for ODH upstream

crobby:
Hmm, it looks like the red-hat-data-services/odh-manifests repo does not have this overlay yet.....which brings up some other process-related questions about things going forward.

crobby:
I've merged the requisite manifest changes and am currently testing this out locally.

crobby:
This is working for me.  I think I saw the Login with OpenShift intermediate page, but that would be unrelated to this PR (would be a manifest change).

PR #73, opened 2021-03-30
Create image for blackbox monitoring of jupyterhub-db

anishasthana:

### DIFF
--- a/monitoring/jupyterhub-db-probe/src/main.py
+++ b/monitoring/jupyterhub-db-probe/src/main.py
@@ -1,0 +1,34 @@
+import os
+import time
+
+from prometheus_client import start_http_server, Gauge
+from sqlalchemy import create_engine
+
+DATABASE_RESPONSE_TIME = Gauge('jupyterhub_db_response_time', 'Time taken for Jupyterhub DB to respond. Negative values indicate failures')
+
+def main():
+    user = os.getenv("JUPYTERHUB_DB_USER", "jupyterhub")
+    password = os.getenv("JUPYTERHUB_DB_PASSWORD", "secretpassword")
+    host = 'jupyterhub-db.redhat-ods-applications.svc.cluster.local'
+    port = 5432
+
+    start_http_server(8080)
+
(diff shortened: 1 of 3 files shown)


monitoring/jupyterhub-db-probe/src/main.py
```diff
@@ -0,0 +1,33 @@
+import os
+import time
+
+from prometheus_client import start_http_server, Gauge
+from sqlalchemy import create_engine
+
+DATABASE_RESPONSE_TIME = Gauge('jupyterhub_db_response_time', 'Time taken for Jupyterhub DB to respond. Negative values indicate failures')
+
+def main():
+    user = os.getenv("JUPYTERHUB_DB_USER", "jupyterhub")
```
maulikjs, line 10: should we keep defaults here?

monitoring/jupyterhub-db-probe/src/main.py
```diff
@@ -0,0 +1,33 @@
...
+    while True:
+        DATABASE_RESPONSE_TIME.set(connect_to_db(user, password, host, port))
+
+
+def connect_to_db(user, password, host, port):
+    try:
+        start_time = time.time_ns() // 1_000_000
+        db_string = f"postgresql://{user}:{password}@{host}:{port}/jupyterhub"
+        db = create_engine(db_string)
+        foo = db.execute("SELECT * FROM users")
```
maulikjs, line 26: I'd say add a `LIMIT 1` to this query for consistency

monitoring/jupyterhub-db-probe/src/main.py
anishasthana: I like having the defaults cause there is some stuff in flux upstream for ODH and secrets. So while that stuff is being done we have something working. I'm happy to remove it once the secrets stuff is set in stone.

maulikjs: sounds good 👍

maulikjs (approved):

PR #77, opened 2021-04-05
Add default groups configuration

vpavlin:


vpavlin:
@crobby @accorvin  WDYT?

deploy.sh
```diff
@@ -129,3 +129,12 @@ cluster_domain=$(oc get ingresses.config.openshift.io cluster --template {{.spec
 odh_dashboard_route="https://odh-dashboard-$ODH_PROJECT.$cluster_domain"
 sed -i "s#<rhods-dashboard-url>#$odh_dashboard_route#g" consolelink/consolelink.yaml
 oc apply -f consolelink/consolelink.yaml
+
+group_cm="configmap/rhods-groups-config"
+exists=$(oc get -n $ODH_PROJECT ${group_cm} -o name | grep ${group_cm} || echo "false")
+unmodified=$(oc get -n $ODH_PROJECT cm -l "opendatahub.io/modified=false" -o name)
```
accorvin, line 135: Should this include a grep for $group_cm? Is it possible for multiple configmaps to exist in the namespace with this label?

Dockerfile
```diff
@@ -22,10 +22,12 @@ COPY deploy.sh $HOME
 COPY opendatahub.yaml $HOME
 ADD monitoring $HOME/monitoring
 ADD consolelink $HOME/consolelink
+ADD groups $HOME/groups
```
accorvin, line 25: I'm unclear what the purpose of this line is. Should there be a file called `groups` in the repo?

vpavlin: The CM is located in `groups` folder

deploy.sh
vpavlin: Good point..at this point we would expect only this one, but in the future, potentially..I can add it, or we can add it if you want

accorvin: I think we should add it in now. This feels like something we would miss in the future and would cause us to overwrite someone's changes.

crobby:
> @crobby @accorvin WDYT?

Hoping to try this today.
Once we have this in, does it become mandatory that when setting up the IdP stuff in OSD that the auth that gets set up has these groups?  ie:  How broken would QE be if we don't get them this info ahead of time?

deploy.sh
vpavlin: @accorvin I pulled the check into a function and made it a bit more general, so it can be used  for other resources. PTAL

vpavlin:
> How broken would QE be if we don't get them this info ahead of time?

They wouldn't be able to login to JH, so pretty broken.

Quick fix for them, before they have the right groups etc in place, can be to just se the `modified` label to `true` and leave the groups empty - that would make it work like it works in upstream ODH

deploy.sh
accorvin: I like it.

PR #78, opened 2021-04-05
Add slo alerts

anishasthana:
cc @hemajv

### DIFF
--- a/monitoring/jupyterhub-db-probe/src/main.py
+++ b/monitoring/jupyterhub-db-probe/src/main.py
@@ -4,7 +4,7 @@ import time
 from prometheus_client import start_http_server, Gauge
 from sqlalchemy import create_engine
 
-DATABASE_RESPONSE_TIME = Gauge('jupyterhub_db_response_time', 'Time taken for Jupyterhub DB to respond. Negative values indicate failures')
+DATABASE_PROBE_SUCCESS = Gauge('jupyterhub_db_probe_success', 'Whether the Jupyterhub DB probe succeded. 1 indicates a success')
 
 def main():
     user = os.getenv("JUPYTERHUB_DB_USER", "jupyterhub")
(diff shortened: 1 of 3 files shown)


anishasthana:
I'd like reviews from all 3 of you on this PR. I'd recommend first looking at this PR commit by commit, and then as a whole. Please review this one carefully, there's a ton of yaml and promql here, and I'd hate for some logic error to slip by.

monitoring/jupyterhub-db-probe/src/main.py
```diff
@@ -15,17 +14,16 @@ def main():
...
 def connect_to_db(user, password, host, port):
     try:
-        start_time = time.time_ns() // 1_000_000
         db_string = f"postgresql://{user}:{password}@{host}:{port}/jupyterhub"
         db = create_engine(db_string)
         foo = db.execute("SELECT * FROM users LIMIT 1")
-        return ((time.time_ns() // 1_000_000) - start_time)
+        return (1)
     except:
         return (-1)
```
4n4nd, line 28: @anishasthana I think this should be returning a `0` instead of `-1`, I think this is how a "down" service is usually denoted in Prom metrics.

4n4nd (changes requested):
one small change, but other than that looks good to me :+1:

monitoring/jupyterhub-db-probe/src/main.py
anishasthana: Good catch. I forgot to update this when I updated the other expression

monitoring/prometheus/prometheus.yaml
```diff
@@ -215,29 +422,111 @@ data:
...
-      - name: RHODS Availability
-        interval: 10m
+        - alert: RHODS Route Error Burn Rate
+          annotations:
+            message: 'High error budget burn for {{ $labels.route }} (current value: {{ $value }})'
+          expr: |
+            sum(haproxy_backend_http_responses_total:burnrate5m{route=~"jupyterhub|odh-dashboard"}) > (14.40 * (1-0.98000))
+            and
+            sum(haproxy_backend_http_responses_total:burnrate1h{route=~"jupyterhub|odh-dashboard"}) > (14.40 * (1-0.98000))
+          for: 2m
```
hemajv, line 435: just curious , what was the reasoning behind the various `expr` combinations? i.e. alerting on burn rates of `5m and 1h`, `30m and 6h` etc and how are the thresholds determined?

anishasthana: Based it on https://promtools.dev/alerts/ and the Google SRE book. The slo promtool was super valuable in going from the tables in the book to actual promql

hemajv: ah I see, the slo promtool looks pretty neat! :smiley:

hemajv:
lgtm! :ship:

maulikjs (approved):

PR #79, opened 2021-04-07
[WIP] Link reader secret to default SA if it exists

tmckayus:


tmckayus:
This works, but the secret name might change so holding for now ....

PR #84, opened 2021-04-12
Add label to generated namespaces

VaishnaviHire:
This commit adds `redhat-openshift-data-science/generated-namespace=true`
label to the namspaces created by odh-deployer. This is required so that odh
operator can acces these namespaces during uninstall using the labels instead of
hardcoding the names.


maulikjs (approved):

deploy.sh
```diff
@@ -69,6 +69,7 @@ function oc::object::safe::to::apply() {
 ODH_PROJECT=${ODH_CR_NAMESPACE:-"redhat-ods-applications"}
 ODH_MONITORING_PROJECT=${ODH_MONITORING_NAMESPACE:-"redhat-ods-monitoring"}
 oc new-project ${ODH_PROJECT} || echo "INFO: ${ODH_PROJECT} project already exists."
+oc label namespace $ODH_PROJECT redhat-openshift-data-science/generated-namespace=true --overwrite=true
```
crobby, line 72: Any chance of this ever failing (even on a cluster update)?  If so, and if that failure might just be due to these things already existing, we might want to add an || "echo <something about what happened and that it's ok>

maulikjs, line 72: actually, should we be adding the label part in quotes?  ""

VaishnaviHire: I think it should work, given that we are adding monitoring label in similar way https://github.com/red-hat-data-services/odh-deployer/blob/main/deploy.sh#L96

crobby (approved):

PR #88, opened 2021-04-19
Add Anaconda CE manifest and secret

Xaenalt:


opendatahub.yaml
```diff
@@ -41,6 +41,11 @@ spec:
         name: manifests
         path: odh-dashboard
     name: odh-dashboard
+  - kustomizeConfig:
+      repoRef:
+        name: manifests
+        path: partners/anaconda
+    name: odh-dashboard
```
cfchase, line 48: I don't think this should have the same name as the odh-dashboard

cfchase (changes requested):

crobby (changes requested):
You're going to need to change the Dockerfile to include the "partners" subdir

Xaenalt:
Added

crobby (approved):
lgtm

anishasthana (changes requested):
@Xaenalt Can you squash the commits?

anishasthana (approved):
/lgtm

cfchase (approved):

crobby (approved):
I have tested this along with the manifest change and the dashboard change.  I think we're in good shape.

PR #96, opened 2021-05-25
Add resource requirements

anishasthana:


monitoring/prometheus/prometheus.yaml
```diff
@@ -72,7 +72,13 @@ spec:
         ports:
         - containerPort: 9090
           name: http
-        resources: {}
+        resources:
+          limits:
+            cpu: 2000m
```
maulikjs, line 77: can you change this to 2 instead of 2000m?

monitoring/prometheus/prometheus.yaml
```diff
@@ -72,7 +72,13 @@ spec:
...
+            memory: 2Gi
+          requests:
+            cpu: 1000m
```
maulikjs, line 80: can you change this to 1 instead of 1000m?

harshad16 (commented):
/lgtm

if the cluster that are provided to the users, has these resource then it is cool :+1:

crobby:
1 whole cpu just for prometheus monitoring seems high....Is that approximately what it actually needs?

crobby:
In my instance, which isn't doing much, prometheus appears to use about 3.3G of memory.  We may require a higher limit.
Screenshot from 2021-05-25 06-24-43

maulikjs (approved):

PR #112, opened 2021-06-11
Add alerting rules for DeadManSnitch

maulikjs:


monitoring/prometheus/prometheus.yaml
```diff
@@ -430,6 +430,17 @@ data:
...
 
+      - name: DeadManSnitch
+        interval: 1m
+        rules:
+          - alert: DeadManSwitch
+            expr: vector(1)
+            labels:
+              severity: critical
+            annotations:
+              description: RHODS monitoring and alerting pipeline is offline.
```
anishasthana, line 441: It should be the other way around right? The description of this alert is that the pipeline _is_ online.

anishasthana (changes requested):

monitoring/prometheus/prometheus.yaml
maulikjs: ah right yes

maulikjs: updated

anishasthana (approved):
Changes look good, but I'd like for us to actually test it out. Do you have a deployment on stage OSD we can look at?
Can you also make a corresponding PR to https://gitlab.cee.redhat.com/croberts/olminstall for the dms secret creation so that we can have an easy way to continue installing RHODS?

maulikjs:
@anishasthana the secret is created as part of the add-on installation workflow. REF: https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/889

anishasthana:
Right, but can you create a dummy secret so we can test in non-OSD clusters internally.

maulikjs:
Ah ok raising a MR then

crobby (approved):
seems to work now, lgtm

crobby (approved):
Seems to be working

PR #115, opened 2021-06-17
Increase prometheus memory limit to 2Gi

accorvin:
The prometheus container was getting OOMKilled. This doubles its memory
limit.

### DIFF
--- a/monitoring/prometheus/prometheus.yaml
+++ b/monitoring/prometheus/prometheus.yaml
@@ -75,10 +75,10 @@ spec:
         resources:
           limits:
             cpu: 400m
-            memory: 3500Mi
+            memory: 4Gi
           requests:
             cpu: 200m
-            memory: 1Gi
+            memory: 2Gi
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         volumeMounts:


maulikjs (approved):

gmfrasca (commented):
/lgtm

accorvin:
We've been having a conversation in Slack about what the appropriate request and limits should be. @maulikjs I believe you said you've seen this spike up as high as ~3 Gi usage, right?

monitoring/prometheus/prometheus.yaml
```diff
@@ -78,7 +78,7 @@ spec:
             memory: 3500Mi
```
maulikjs, line 78: @accorvin so I have seen it spike to around 3200Mi which led me to bump this to 3500Mi.
We should bump the limit 4Gi and keep the requests at 2Gi?

maulikjs: Just so we can give it a bit more breathing space

accorvin: Seems reasonable. Will update.

maulikjs (approved):

PR #123, opened 2021-07-13
Set tags for cuda buildchain repositories

tmckayus:
Add tags of the form "nb-N" for the rest of the cuda build chain
repositories in red-hat-data-services.  We did this for the
base images, add tags for minimal, pytorch, and tensorflow as
well so that the notebook source does not change for a release
because of a commit to the repository.


tmckayus:
@harshad16 @vpavlin ptal

This change does not modify the current state, just replaces the python38 branch names with a new tag (added to the corresponding repos)

harshad16 (approved):
lgtm , thanks :100:

PR #132, opened 2021-07-30
Update the ref for s2i-thoth in the cuda build chain

tmckayus:
Changed to tag nb-2


tmckayus:
Corresponds to this tag (curl removal)

https://github.com/red-hat-data-services/s2i-thoth/tree/nb-2

anishasthana (approved):

maulikjs (approved):

PR #139, opened 2021-08-10
Remove CUDA builds

maulikjs:
https://issues.redhat.com/browse/RHODS-1651


crobby (approved):

anishasthana:
Aren't the pytorch and tensorflow notebooks that are built here part of the images we _always_ want?

anishasthana (changes requested):
Don't we need to remove https://github.com/red-hat-data-services/odh-deployer/blob/main/deploy.sh#L101?

LaVLaS:
> Aren't the pytorch and tensorflow notebooks that are built here part of the images we _always_ want?

This is a good point. This PR should be paired with another PR to deploy the non-gpu builds for pytorch and tensorflow.  Since we are using the thoth builds for minimal and generic-data-science, is the same pre-built notebook exception allowed  for these non-gpu notebooks?

crobby (commented):
I think we just want to remove the build and imagestream for gpu-minimal (cuda)

harshad16:
> Don't we need to remove https://github.com/red-hat-data-services/odh-deployer/blob/main/deploy.sh#L101?

we should remove this as well, as it would break the deployer , if the file is missing.

crobby (changes requested):
So, the newly requested funcitonality via RHODS-1651 is to just remove the annotation from the cuda imagestream so that it will not appear in the spawner UI

anishasthana:
No longer needed.

PR #142, opened 2021-08-11
Add smtp config to send user notifications and add the corresponding altermanager router

maulikjs:


deploy.sh
```diff
@@ -70,6 +70,7 @@ ODH_PROJECT=${ODH_CR_NAMESPACE:-"redhat-ods-applications"}
 ODH_MONITORING_PROJECT=${ODH_MONITORING_NAMESPACE:-"redhat-ods-monitoring"}
 ODH_NOTEBOOK_PROJECT=${ODH_NOTEBOOK_NAMESPACE:-"rhods-notebooks"}
 CRO_PROJECT=${CRO_NAMESPACE:-"redhat-ods-operator"}
+ODH_OPERATOR_PROJECT="redhat-ods-operator"
```
crobby, line 73: Might as well make this look like the other namespace definitions and have an env var with a default value.  Not sure we'll ever really use them, but it would be consistent

crobby:
My deployer just sort of gets stuck at:  Error from server (NotFound): secrets "redhat-rhods-smtp" not found.  Is there another PR somewhere to include that secret?

crobby:
> My deployer just sort of gets stuck at: Error from server (NotFound): secrets "redhat-rhods-smtp" not found. Is there another PR somewhere to include that secret?

I should clarify....the deployer doesn't crash, it does eventually continue, but I don't think it ever finds that secret.
It also seems to be looking the secret "addon-managed-odh-parameters" without success

Are those supposed to come from the managed-tenants stuff?

crobby (approved):
appears to be working

PR #145, opened 2021-08-12
Change the context and dockerpath for thoth buildconfig

tmckayus:
This is related to https://github.com/red-hat-data-services/s2i-thoth/pull/7
which removed instances of curl from the Dockerfile


tmckayus:
Verified on a real install that this change fixed an existing broken build config.
It was part of https://github.com/red-hat-data-services/s2i-thoth/pull/7 but was left out.

@crobby @anishasthana

PR #148, opened 2021-08-17
RBAC for dedicated-admins in select namespaces if installed on OSD

gmfrasca:
- Grant CRUD for CMs, ISs, BCs, Builds and Secrets on the applications and JH notebooks namespaces if install target is OpenShift Dedicated
- dedicated-admins group does not exist if not on OSD, so don't install otherwise

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ COPY opendatahub.yaml $HOME
 COPY opendatahub-osd.yaml $HOME
 COPY rhods-monitoring.yaml $HOME
 COPY rhods-notebooks.yaml $HOME
+COPY rhods-osd-configs.yaml $HOME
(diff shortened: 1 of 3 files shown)


anishasthana (changes requested):
Can you link to the related JIRA issue in the PR?

gmfrasca:
PR: https://issues.redhat.com/browse/RHODS-1590

anishasthana (commented):
Is this deployed anywhere/ available as a built image?

anishasthana (approved):
I have verified that the roles and rolebindings are created. They also granted enough permissions to dedicated-admins that we were expecting. 

This should only be merged in after the changes in https://github.com/red-hat-data-services/odh-manifests/pull/144 have made their way into the opendatahub-operator repo.

PR #152, opened 2021-08-23
Add init container in blackbox-exporter to wait for RHODS services to become online

rimolive:


anishasthana (commented):
Changes look good. We could possible bump up seq from 150 to 200 -- lets see how it goes in the test

monitoring/prometheus/blackbox-exporter-common.yaml
```diff
@@ -55,6 +55,13 @@ spec:
       - name: config-volume
         configMap:
           name: blackbox
+      initContainers:
+        - name: wait-for-deployment
+          image: 'registry.access.redhat.com/ubi8/ubi-minimal:8.4-208'
+          command:
+            - /bin/sh
+            - '-c'
+            - for i in `seq 1 150`; do sleep 10; echo "Waiting for odh-dashboard service to become available..."; if curl -s -k https://odh-dashboard.redhat-ods-applications.svc:8443; then exit 0; fi; done; exit 1
```
anishasthana, line 64: Actually. @rimolive can we change this to 230 from 150?

anishasthana (approved):
image

The alerts are no longer firing. This is behaving as expected.

PR #189, opened 2021-11-03
Add relabeling of build objects in buildchain.sh

tmckayus:
When the version doesn't match but the checksum does, build objects
should be relabeled with the current version along with buildconfigs
and imagestreams (which are already happening).

- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-1990
- [x] The Jira story is acked

### DIFF
--- a/buildchain.sh
+++ b/buildchain.sh
@@ -48,6 +48,11 @@ elif [ "$version" == "true" ]; then
         oc label buildconfig -n $ODH_PROJECT "$name" rhods/buildchain=cuda-"$RHODS_VERSION" --overwrite=true
     done
 
+    bu=($(oc get build -l rhods/buildchain -n $ODH_PROJECT -o jsonpath=' {range .items [*]} {.metadata.name}'))
+    for name in ${bu[*]}; do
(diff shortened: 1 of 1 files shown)


buildchain.sh
```diff
@@ -48,6 +48,11 @@ elif [ "$version" == "true" ]; then
         oc label buildconfig -n $ODH_PROJECT "$name" rhods/buildchain=cuda-"$RHODS_VERSION" --overwrite=true
     done
 
+    bu=($(oc get build -l rhods/buildchain -n $ODH_PROJECT -o jsonpath=' {range .items [*]} {.metadata.name}'))
```
anishasthana, line 51: What does `bu` stand for?

tmckayus: "bu", it's just a bash variable :)

tmckayus: @anishasthana would you like bd, or bld, or something else? :) Happy to oblige

crobby (approved):
Ok, I finally was able to try this out.  lgtm

PR #192, opened 2021-11-22
Change imagestream tags to py3.8 for generic and minimal notebooks

tmckayus:
Change imagestream tags for the generic data science notebook and minimal notebook to py3.8 to create a stable, meaningful tag following the pattern of "py3.8-cuda-11.0.3" on the runtime buildchain. This tag based on the Python version will allow us to update the actual image referenced by the imagestream without having the image version explicitly in the tag, which leads to tag management issues on upgrades.

As a one-time fix, the deployer will retag existing v0.0.5 and v0.0.15 images with the py3.8 tag and then delete the v0.0.5 and v0.0.15 tags.

- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-973
- [x] The Jira story is acked 

### DIFF
--- a/buildchain.sh
+++ b/buildchain.sh
@@ -4,6 +4,7 @@ ODH_PROJECT=${ODH_CR_NAMESPACE:-"redhat-ods-applications"}
 checksum=false
 version=false
 
+# Figure out if the RHODS version has changed or the buildchain manifest checksum
 res=$(oc get cm rhods-buildchain -n $ODH_PROJECT)
 if [ "$?" -eq 0 ]; then
     # See if the version matches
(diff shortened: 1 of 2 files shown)


tmckayus:
As part of RHODS-973, it's necessary to use the red-hat-data-services repository as the source for the build of the generic notebook.  As part of that, we should have a significant version tag that we can use to track the notebook version separate from upstream thoth-station. Since RHODS-1990 calls for obvious end user cues on upgrade that a successul upgrade has happened, we will use the RHODS version as the tag on the generic data science and minimal python notebook imagestreams.

However, if the operator simply applies an updated imagestream with a different tag, it will leave existing tags and simply add the new one. This will lead to a build up of versions across installs. Consequently, if the version is different (indicated by the rhods-buildchain configmap) on startup, the odh-deployer init container will delete the imagestreams for prebuilt images and they will be recreated by the operator with a single tag.

tmckayus:
Related to https://github.com/red-hat-data-services/odh-manifests/pull/186 where the manifests are updated to use the py3.8 tag (tags will have been moved by the deployer)

crobby (approved):
I've seen this work, lgtm

PR #216, opened 2022-03-09
Add culler timeout value and add jh-idle-culler overlay

maroroman:
- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-2324
- [x] The Jira story is acked
- [ ] An entry has been added to the latest build document in [Build Announcements Folder](https://drive.google.com/drive/folders/1sgkK1WZgGo9CXsLizNe0GbAzVKuSKrZL).

Testing instructions:
`jupyterhub-cfg` Configmap should be deploy with the `culler_timeout` value in `data`
`jupyterhub-idle-culler` deployment config should be present and deployed.

### DIFF
--- a/jupyterhub/jupyterhub-configmap.yaml
+++ b/jupyterhub/jupyterhub-configmap.yaml
@@ -11,3 +11,4 @@ data:
   gpu_mode: ""
   singleuser_pvc_size: 20Gi
   notebook_destination: <notebook_destination>
+  culler_timeout: "31536000" # 1 year
--- a/opendatahub-osd.yaml
+++ b/opendatahub-osd.yaml
@@ -23,6 +23,7 @@ spec:
       overlays:
         - jupyterhub-rds-db-patches
         - ui-config
+        - ../overlays/jupyterhub-idle-culler
       repoRef:
         name: manifests
         path: jupyterhub/jupyterhub/overrides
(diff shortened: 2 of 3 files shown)


jupyterhub/jupyterhub-configmap.yaml
```diff
@@ -11,3 +11,4 @@ data:
   gpu_mode: ""
   singleuser_pvc_size: 20Gi
   notebook_destination: <notebook_destination>
+  culler_timeout: "31536000 " # 1 year
```
LaVLaS, line 14: Should the default timeout be 1 year when it was previously 3 days?

maroroman: As per comments here: https://issues.redhat.com/browse/RHODS-2324

The default cannot be set to infinity (no culling) so we decided to provide a high starting value for "no culling" so that notebooks are culled only if the administrator wants them to and the culler is enabled by default.

crobby (changes requested):
I will try and get a live build together for this, but I don't think it will work.  The KfDef, I think, will need to look like the change tried out here:  https://github.com/crobby/odh-deployer/commit/8dfd193b3115fb8fa47ed07e2a3e4abcc9e6cd9b

crobby:
Obviated by:  https://github.com/red-hat-data-services/odh-deployer/pull/224

crobby:
Needs a rebase

crobby (approved):

maroroman:
Replaced by #228

PR #227, opened 2022-04-05
Update OSD check to use an OpenShift configmap

anishasthana:

- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-3454
- [x] The Jira story is acked

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -101,11 +101,12 @@ oc create -n ${ODH_PROJECT} -f monitoring/jupyterhub-prometheus-token-secrets.ya
 
 sed -i "s/<notebook_destination>/$ODH_NOTEBOOK_PROJECT/g" jupyterhub/jupyterhub-configmap.yaml
 oc apply -n ${ODH_PROJECT} -f jupyterhub/jupyterhub-configmap.yaml || echo "INFO: Jupyterhub ConfigMap already created "
-# Check if the installation target is OSD to determine the deployment manifest path
-deploy_on_osd=0
-oc get group dedicated-admins &> /dev/null || deploy_on_osd=1
-if [ "$deploy_on_osd" -eq 0 ]; then
-  # On OpenShift Dedicated, deploy with CRO
+
+infrastructure=$(oc get infrastructure cluster -o jsonpath='{.spec.platformSpec.type}')
+# Check if the installation target is AWS to determine the deployment manifest path
+if [ $infrastructure = "AWS" ]; then
+  # On AWS OpenShift Dedicated, deploy with CRO
+  echo "INFO: Deploying on AWS. Creating CRO for deployment of RDS Instance"
   ODH_MANIFESTS="opendatahub-osd.yaml"
 
   # Install CRO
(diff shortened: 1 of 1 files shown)


anishasthana:
@pablofelix  Live build is available at quay.io/anishasthana/rhods-operator-live-catalog:1.90.0-3454. It's upgrading from `1.8.0-14` as there isn't an official 1.9.0 build yet.

deploy.sh
```diff
@@ -101,11 +101,10 @@ oc create -n ${ODH_PROJECT} -f monitoring/jupyterhub-prometheus-token-secrets.ya
...
 oc apply -n ${ODH_PROJECT} -f jupyterhub/jupyterhub-configmap.yaml || echo "INFO: Jupyterhub ConfigMap already created "
-# Check if the installation target is OSD to determine the deployment manifest path
-deploy_on_osd=0
-oc get group dedicated-admins &> /dev/null || deploy_on_osd=1
-if [ "$deploy_on_osd" -eq 0 ]; then
-  # On OpenShift Dedicated, deploy with CRO
+
+# Check if the installation target is AWS to determine the deployment manifest path
+if [ "$(oc get infrastructure cluster -o jsonpath='{.spec.platformSpec.type}')" = "AWS" ]; then
+  # On AWS OpenShift Dedicated, deploy with CRO
```
crobby, line 107: Can we maybe log a message here of some sort so it can be obvious?

crobby (changes requested):
I think a log noting which env, AWS or "not", would help....maybe bonus points for the log entry indicating what else that will mean (the AWS DB getting created instead of local)

crobby (approved):
/lgtm

anishasthana:
quay.io/anishasthana/rhods-operator-live-catalog:1.92.0-3454 is the livebuild with the additional logging, in case anyone is interested!

deploy.sh
```diff
@@ -101,11 +101,11 @@ oc create -n ${ODH_PROJECT} -f monitoring/jupyterhub-prometheus-token-secrets.ya
...
 sed -i "s/<notebook_destination>/$ODH_NOTEBOOK_PROJECT/g" jupyterhub/jupyterhub-configmap.yaml
 oc apply -n ${ODH_PROJECT} -f jupyterhub/jupyterhub-configmap.yaml || echo "INFO: Jupyterhub ConfigMap already created "
-# Check if the installation target is OSD to determine the deployment manifest path
-deploy_on_osd=0
-oc get group dedicated-admins &> /dev/null || deploy_on_osd=1
-if [ "$deploy_on_osd" -eq 0 ]; then
-  # On OpenShift Dedicated, deploy with CRO
+
+# Check if the installation target is AWS to determine the deployment manifest path
+if [ "$(oc get infrastructure cluster -o jsonpath='{.spec.platformSpec.type}')" = "AWS" ]; then
```
accorvin, line 106: I worry that this, as it, might eventually bite us when we look to deploy on other cloud providers. Can we update this to somehow fail if we're running on something other than PSI or AWS?

anishasthana: (it's not a hard check to implement) I'm not sure it's actually worth it. The way I see it, right now AWS is a special case with requiring CRO, and PSI is a "base" case. If we need to deploy on GCP, or ARO, I imagine it'd be as simple as specifying an additional if-else.

accorvin: What I'm saying is that if we forget to come back and add that additional if-else (seems like a very easy thing to forget) then this will get deployed in GCP without HA.

LaVLaS: > What I'm saying is that if we forget to come back and add that additional if-else (seems like a very easy thing to forget) then this will get deployed in GCP without HA.

In that scenario, wouldn't HA support be part of the GCP "on-boarding" development process?  If it fails then we would "un-break" RHODS on GCP before we could start working on replacing the postgres `jupyterhub-db`

accorvin: Yes that's true. I'm probably being overly cautious in this. I'll defer to you all if you're good with this plan.

anishasthana: Bah. I see your point. Updated.

anishasthana:
Hopefully final live build: quay.io/anishasthana/rhods-operator-live-catalog:1.97.0-3454

accorvin:
I'm spinning up a ROSA cluster to test this on now.

accorvin:
I see this line in the odh-deployer pod on my ROSA cluster: ```INFO: Deploying on AWS. Creating CRO for deployment of RDS Instance```
So I think it's good.

accorvin:
Another update: I see the following on a PSI OSIA cluster:
```
INFO: Deploying on PSI. Creating local database
```

I can confirm that the CRO pod does not get created in PSI, and the PostgreSQL pod does.

So, everything looks good on both PSI and ROSA.

lucferbux (approved):
/lgtm

lucferbux:
Checked on PSI, works great.

PR #234, opened 2022-06-13
Update Prometheus metrics

VaishnaviHire:
- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] For commits that came from upstream, `[UPSTREAM]` has been prepended to the commit message
- [ ] JIRA link(s): https://issues.redhat.com/browse/RHODS-4071
- [ ] The Jira story is acked


VaishnaviHire:
The changes include:
1. Moving Dashboard `probe_success` alert to `warning`
     - The Dashboard `Route Error Burn Rate` will be used for critical alerts.
2. Moving traefik `probe_success` alert to `warning`
     - Users can access jupyterhub as long as 1 traefik pod is running
     -  When all the traefik pods are down, JH `probe_success` metric is also triggered as shown in the graph below
     - We would be using JH `probe_success` metric to notify if all traefik pods are down
     - Screen Shot 2022-06-13 at 3 09 58 PM

VaishnaviHire:
Working on the live-build

VaishnaviHire:
@anishasthana @cfchase Can you please review this PR?

anishasthana:
Cc @jgarciao

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -290,35 +290,59 @@ data:
...
-        - alert: RHODS Probe Success Burn Rate
+        - alert: RHODS JupyterHub Probe Success Burn Rate
           annotations:
             message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
-            summary: RHODS Route Error Burn Rate
+            summary: RHODS JupyterHub Probe Success Burn Rate
           expr: |
-            sum(probe_success:burnrate30m{instance=~"jupyterhub|rhods-dashboard|jupyterhub-db|traefik"}) by (instance) > (6.00 * (1-0.99950))
+            sum(probe_success:burnrate30m{instance=~"jupyterhub|jupyterhub-db"}) by (instance) > (6.00 * (1-0.99950))
```
cfchase, line 311: @anishasthana I'm trying to grok these tests and I'm new to these alerts and promql.  It looks like we're taking a `sum` on the left hand side of 30 minutes `burnrate30m` and an 6 hours `burnrate6h`, which I would think would be very different, and comparing it to the same value `(6.00 * (1-0.99950))` (0.003) on the right hand side, which seems odd to me.  I'm probably missing something but I would have expected to take the average burn rate over that time period  OR seeing a different value on the right hand side.  I'm sure I'm misreading something, but I would like to make sure we're not burning our error rate for some other reason we're not fixing.

VaishnaviHire:
Live Build: quay.io/modh/rhods-operator-live-catalog:1.13.0-rhods-4071 

Testing steps for current changes:
1. Verify if the new `critical` alerts are defined as `RHODS JupyterHub Probe Success Burn Rate` 
2. Verify that the `RHODS Route Error Burn Rate` is updated to `RHODS Dashboard Route Error Burn Rate`
3.  Scale down `traefik` deployment and observe if it eventually triggers `RHODS JupyterHub Probe Success Burn Rate`  
4. Scale down `rhods-dashboard` deployment and observe if it eventually triggers `RHODS Dashboard Route Error Burn Rate`  

cc: @jgarciao @cfchase

monitoring/prometheus/prometheus-configs.yaml
Ajpantuso: See the [SRE workbook](https://sre.google/workbook/alerting-on-slos/#6-multiwindow-multi-burn-rate-alerts) The idea is to only alert when the error budget is being actively consumed within the longer window (to localize the alert). There are also handy [generators](https://promtools.dev/alerts/errors) for this.

VaishnaviHire: @Ajpantuso  I think we would like to confirm if we should be comparing both the smaller and longer window with the same value((6.00 * (1-0.99950))), given that we are taking the `sum` of the burnrate.

Ajpantuso: That outer sum probably came from [slo-libsonnet](https://github.com/metalmatze/slo-libsonnet/blob/master/slo-libsonnet/error-burn.libsonnet#L58-L60) or something referring to it. In this case `sum(vector) by (instance)` appears to be equivalent to `vector` since there are no other dimensions which have multiple values per `instance`. So you can read `sum(probe_success:burnrate30m{instance=~"jupyterhub|jupyterhub-db"}) by (instance) > (6.00 * (1-0.99950))` as `probe_success:burnrate30m{instance=~"jupyterhub|jupyterhub-db"} > (6.00 * (1-0.99950))`. And for additional clarity `sum_over_time` is the operator used to sum over a range of time-series values rather than sum over a range of dimension values.

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -290,35 +290,59 @@ data:
...
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
-            summary: RHODS Route Error Burn Rate
+            summary: RHODS Probe Success Burn Rate
+          expr: |
+            sum(probe_success:burnrate5m{instance=~"rhods-dashboard|traefik"}) by (instance) > (14.40 * (1-0.99950))
+            and
+            sum(probe_success:burnrate1h{instance=~"rhods-dashboard|traefik"}) by (instance) > (14.40 * (1-0.99950))
+          for: 2m
+          labels:
+            severity: warning
```
Ajpantuso, line 328:
```suggestion
            severity: critical
```
I think we may want to leave this as `critical`. From Pagerduty I see only 3 alerts generated from this highest degree of burnrate for SRE and only 1 of those auto-resolved as an anomaly. We can reduce this to `warning` later, but at that point we may have to re-evaluate the SLO itself if we are no longer concerned about error budget.

cfchase (approved):

monitoring/prometheus/prometheus-configs.yaml
VaishnaviHire: We moved this to `warning` because we have a `Route Burn Rate Alert` for Dashboard as well that is `critical`

VaishnaviHire: Here is the alert : https://github.com/VaishnaviHire/odh-deployer/blob/rhods_4071/monitoring/prometheus/prometheus-configs.yaml#L230

Ajpantuso: Cool, then `warning` here is okay with me.

anishasthana (approved):

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -243,20 +243,20 @@ data:
         - alert: RHODS Route Error Burn Rate
```
jgarciao, line 243: I think it should be `RHODS Dashboard Route Error Burn Rate`

monitoring/prometheus/prometheus-configs.yaml
jgarciao: >That outer sum probably came from [slo-libsonnet](https://github.com/metalmatze/slo-libsonnet/blob/master/slo-libsonnet/error-burn.libsonnet#L58-L60) or something referring to it.

Yes, I think the original rules were generated using this online tool: https://promtools.dev/alerts/errors

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -290,35 +290,59 @@ data:
...
+          for: 2m
+          labels:
+            severity: warning
+        - alert: RHODS Probe Success Burn Rate
+          annotations:
+            message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
+            triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
+            summary: RHODS Probe Success Burn Rate
+          expr: |
+            sum(probe_success:burnrate30m{instance=~"rhods-dashboard|traefik"}) by (instance) > (6.00 * (1-0.99950))
```
jgarciao, line 335: I think `RHODS Probe Success Burn` Rate should be adapted too "for 1h"  and "for 3h" with similar changes

jgarciao (changes requested):

jgarciao (changes requested):

monitoring/prometheus/prometheus-configs.yaml
VaishnaviHire: Talked offline, Jorge's suggestion is to divide alerts "for: 1h" and "for: 3h" to `instance=~"rhods-dashboard|traefik"` and  `instance=~"jupyterhub|jupyterhub-db"` as well to make them consistent with other alerts

VaishnaviHire: Updated

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -216,47 +216,47 @@ data:
 
       - name: SLOs-haproxy_backend_http_responses_total
         rules:
-        - alert: RHODS Route Error Burn Rate
+        - alert: RHODS Dashbboard Route Error Burn Rate
```
jgarciao, line 219: Typo in Dashboard

PR #235, opened 2022-06-17
Update prometheus rhods_aggregate_availability metric 

atheo89:
Updated the rhods_aggregate_availability metric to include a label indicating the availability of each component over time. 

- [ ] For commits that came from upstream, `[UPSTREAM]` has been prepended to the commit message
- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-4229
- [x] The Jira story is acked

### Live Build

quay.io/repository/modh/rhods-operator-live-catalog:1.13.0-rhods-4229

### Testing Instructions

1. Deploy the live build
2. Kill all the jupyterhub pods
3. Enter prometheus, check that in rhods_aggregate_availability both "combined" and "jupyterhub" vectors are down
4. Kill all dashboard pods
5. Enter prometheus, check that in rhods_aggregate_availability both "combined" and "dashboard" vectors are down

### DIFF
--- a/monitoring/prometheus/prometheus-configs.yaml
+++ b/monitoring/prometheus/prometheus-configs.yaml
@@ -197,7 +197,7 @@ data:
 
       - name: Availability Metrics
         rules:
-        - expr: min(probe_success)
+        - expr: (min(min_over_time(probe_success[10s])) by (instance) or label_replace(min(min_over_time(probe_success[10s])), "instance", "combined", "instance", ".*"))
           record: rhods_aggregate_availability
 
   alerting.rules: |


lucferbux:
@gmfrasca i think we got something, please, check this out to see how it went, thanks!

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -197,7 +197,7 @@ data:
 
       - name: Availability Metrics
         rules:
-        - expr: min(probe_success)
+        - expr: min(probe_success offset 5h) by (instance) or label_replace(min(probe_success offset 5h), "instance", "combined", "", "")
```
gmfrasca, line 200: I don't think you want those `offset` s in there.... they're usually just for testing against historical data

gmfrasca: I haven't tested so perhaps this works, but does having dst_label and dst_regex (the 4th and 5th params for label_replace) replace more than just the original instance label?   Probably moot point now but I wonder if it will have implications if anyone builds on top of this.  so tl,dr: maybe just replace the instance label?

atheo89: Hi Giulio, thanks for the review. I am going to delete offset, but I am not sure if I have to change **instance -> component:** min(probe_success) by (instance) or label_replace(min(probe_success), "**instance**", "combined", "", "") to **component** for example.

gmfrasca: so going of the corresponding ticket, it looks like the aggregated item just needs to be consistent with the 'split' ones, which appear to be using `instance`.  you could probably relabel _all_ of them to rename `instance` to `component`, but that's probably more work than necessary.

gmfrasca (commented):
/lgtm

gmfrasca (approved):

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -495,7 +496,7 @@ data:
           regex: (.+):(\d+)
           target_label: __address__
           replacement: ${1}:8082
-
+
```
anishasthana, line 499: Can you fix this whitespace that was added?

anishasthana (changes requested):

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -478,6 +478,7 @@ data:
           replacement: ${1}:8081
 
     - job_name: 'Traefik Proxy Metrics'
+      scrape_interval: 10s
```
anishasthana, line 481: If we don't care about further improving traefik accuracy why are we updating the scrape interval here?

anishasthana (changes requested):

monitoring/prometheus/prometheus-configs.yaml
atheo89: You are right, maybe I should remove the scape_interval completely from there because it wasn't before. May I proceed to this deletion?

anishasthana (approved):

pablofelix:
lgtm

pablofelix (approved):
lgtm

PR #247, opened 2022-07-27
Fix monitoring stack

anishasthana:
- Update prometheus federation route logic to account for changes in OpenShift 4.11
- Update deploy script to account for deprecation of/bug in 'oc sa get-token' command

## Description

https://issues.redhat.com/browse/RHODS-4739
https://issues.redhat.com/browse/RHODS-4745

## Merge criteria:

- [ ] JIRA link(s):
- [x] Live build image: 
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -177,27 +177,8 @@ sed -i "s/<prometheus_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/promet
 sed -i "s/<alertmanager_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/prometheus/prometheus-secrets.yaml
 oc create -n $ODH_MONITORING_PROJECT -f monitoring/prometheus/prometheus-secrets.yaml || echo "INFO: Prometheus session secrets already exist."
 
-if [ $(oc get clusterversion -o jsonpath='{.items[0].status.desired.version}' | cut -d '.' -f 1) -gt 4 ]; then
-    route="prometheus-k8s-federate"
-    echo "On OpenShift 5 or higher"
-else
-    if [ $(oc get clusterversion -o jsonpath='{.items[0].status.desired.version}' | cut -d '.' -f 1) -eq 4 ]; then
-        if [ $(oc get clusterversion -o jsonpath='{.items[0].status.desired.version}' | cut -d '.' -f 2) -ge 11 ]; then
-            route="prometheus-k8s-federate"
-            echo "On OpenShift 4.11 or higher"
-        else
-            route="prometheus-k8s"
-            echo "Version of OpenShift is older than 4.11"
-        fi
-    else
-        route="prometheus-k8s"
-        echo "Version of OpenShift is older than 4.11"
-    fi
-fi
-ocp_federate_target=$(oc::wait::object::availability "oc get -n openshift-monitoring route $route -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
 
 sed -i "s/<jupyterhub_prometheus_api_token>/$(oc::wait::object::availability "oc get secret -n $ODH_PROJECT jupyterhub-prometheus-token-secrets -o jsonpath='{.data.PROMETHEUS_API_TOKEN}'" 2 30 | tr -d "'"  | base64 --decode)/g" monitoring/prometheus/prometheus-configs.yaml
-sed -i "s/<federate_target>/$ocp_federate_target/g" monitoring/prometheus/prometheus-configs.yaml
 oc apply -n $ODH_MONITORING_PROJECT -f monitoring/prometheus/alertmanager-svc.yaml
 
 alertmanager_host=$(oc::wait::object::availability "oc get route alertmanager -n $ODH_MONITORING_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
--- a/deploy.sh
+++ b/deploy.sh
@@ -177,8 +177,6 @@ sed -i "s/<prometheus_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/promet
 sed -i "s/<alertmanager_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/prometheus/prometheus-secrets.yaml
 oc create -n $ODH_MONITORING_PROJECT -f monitoring/prometheus/prometheus-secrets.yaml || echo "INFO: Prometheus session secrets already exist."
 
-
-prometheus_token=$(oc::wait::object::availability "oc sa -n $ODH_MONITORING_PROJECT get-token prometheus" 2 30)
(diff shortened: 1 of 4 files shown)


anishasthana:
Live build: quay.io/anishasthana/rhods-operator-live-catalog:1.14.8-4739

atheo89:
@anishasthana  The only drawback that i found while I was checking the fix for RHODS-4745, was on Prometheus Dashboard > Targets > Federate Prometheus target was down with Error: server returned HTTP status 503 Service Unavailable (I am using OpenShift 4.10). But when I restarted rhods operator(as Jorge mentioned) the target started properly. The rest works fine.

crobby:
Screenshot from 2022-07-28 08-28-57

I also see the Federate Prometheus targets as "down" (Openshift 4.10.23)

crobby:
Also, what does the behavior look like if you start with 4.10.x, install RHODS, and then upgrade the cluster to 4.11.x?

crobby:
> Also, what does the behavior look like if you start with 4.10.x, install RHODS, and then upgrade the cluster to 4.11.x?

After more thought, I *think* it should work in this scenario.

anishasthana:
New live build: quay.io/anishasthana/rhods-operator-live-catalog:1.14.10-4739

anishasthana:
I've updated the permissions in the live build repo so that the operator can pull the clusterversion object.

anishasthana:
Latest live build: quay.io/anishasthana/rhods-operator-live-catalog:1.14.12-4739

anishasthana:
@atheo89 I fixed the problem so you will no longer require a restart!

anishasthana:
Once merged, we should also follow up with a merge on https://gitlab.cee.redhat.com/service/managed-tenants-bundles/-/merge_requests/573

deploy.sh
```diff
@@ -177,9 +177,26 @@ sed -i "s/<prometheus_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/promet
 sed -i "s/<alertmanager_proxy_secret>/$(openssl rand -hex 32)/g" monitoring/prometheus/prometheus-secrets.yaml
 oc create -n $ODH_MONITORING_PROJECT -f monitoring/prometheus/prometheus-secrets.yaml || echo "INFO: Prometheus session secrets already exist."
 
+prometheus_token=$(oc::wait::object::availability "oc sa -n $ODH_MONITORING_PROJECT get-token prometheus" 10 30)
```
samuelvl, line 180: There is an alternative way to scrape metrics from internal Prometheus [1] which is more resilient (use internal traffic) and secure (do not read secrets from openshift-monitoring):

- You can allow the RHODS Prometheus SA to get metrics from internal Prometheus by binding the following ClusterRole:

```shell
oc adm policy add-cluster-role-to-user cluster-monitoring-view -z rhods-prometheus-serviceaccount
```

- Use internal Thanos service instead of Prometheus public route. The service doesn't depend on the cluster configuration and can be hardcoded. I haven't checked if changed in 4.11 but if it doesn't we can save the `clusterversion` logic.

1. https://www.redhat.com/en/blog/custom-grafana-dashboards-red-hat-openshift-container-platform-4

samuelvl: This is the list of services in 4.11

```
NAME                                       TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                               AGE
alertmanager-main                          ClusterIP   172.30.24.77     <none>        9094/TCP,9092/TCP,9097/TCP            153m
alertmanager-operated                      ClusterIP   None             <none>        9093/TCP,9094/TCP,9094/UDP            153m
cluster-monitoring-operator                ClusterIP   None             <none>        8443/TCP                              165m
configure-alertmanager-operator            ClusterIP   172.30.148.219   <none>        8080/TCP                              143m
configure-alertmanager-operator-registry   ClusterIP   172.30.78.76     <none>        50051/TCP                             144m
kube-state-metrics                         ClusterIP   None             <none>        8443/TCP,9443/TCP                     153m
node-exporter                              ClusterIP   None             <none>        9100/TCP                              153m
openshift-state-metrics                    ClusterIP   None             <none>        8443/TCP,9443/TCP                     153m
prometheus-adapter                         ClusterIP   172.30.103.169   <none>        443/TCP                               153m
prometheus-k8s                             ClusterIP   172.30.134.72    <none>        9091/TCP,9092/TCP                     153m
prometheus-k8s-thanos-sidecar              ClusterIP   None             <none>        10902/TCP                             153m
prometheus-operated                        ClusterIP   None             <none>        9090/TCP,10901/TCP                    153m
prometheus-operator                        ClusterIP   None             <none>        8443/TCP,8080/TCP                     153m
prometheus-operator-admission-webhook      ClusterIP   172.30.252.20    <none>        8443/TCP                              161m
sre-dns-latency-exporter                   ClusterIP   172.30.116.10    <none>        80/TCP                                144m
sre-ebs-iops-reporter                      ClusterIP   172.30.45.155    <none>        80/TCP                                144m
sre-stuck-ebs-vols                         ClusterIP   172.30.13.128    <none>        80/TCP                                144m
telemeter-client                           ClusterIP   None             <none>        8443/TCP                              153m
thanos-querier                             ClusterIP   172.30.27.57     <none>        9091/TCP,9092/TCP,9093/TCP,9094/TCP   153m
token-refresher                            ClusterIP   172.30.108.225   <none>        80/TCP                                144m
``` 

`thanos-querier` and `prometheus-k8s` do not change.

anishasthana: So the reason we went with this was that "federation" is the recommended way for prometheuses to share metrics without overwriting labels and the like.

We're also not reading secrets from the `openshift-monitoring` namespace, but the `redhat-ods-monitorin` namespace.

re: the `cluster-monitoring-view` user, good point. As I recall there was some reason we had ended up manually specifying the token in the federation config.
Our Prometheus already appears to have cluster reader: https://github.com/red-hat-data-services/odh-manifests/blob/master/monitoring/base/cluster-monitor-rbac.yaml

samuelvl: Then we can replace this line:

https://github.com/red-hat-data-services/odh-deployer/blob/87da022c5bf024d65e30aef01f95254c86f86676/monitoring/prometheus/prometheus-configs.yaml#L439

With:

```yaml
bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
```

In this way we avoid putting a secret into a configmap and the need of using the `oc sa get-token`.

samuelvl: `So the reason we went with this was that "federation" is the recommended way for prometheuses to share metrics without overwriting labels and the like.`
Both endpoints expose same data. There is no need to use the external route (LB + HAProxy) when there is an internal service that exposes same information.

samuelvl: New changes looks good! Thanks @anishasthana

anishasthana: Updated!

samuelvl (approved):
/lgtm

PR #250, opened 2022-08-02
WIP: Kfnbc migration

anishasthana:
This is pretty meaty. I would review it commit by commit. The PR is currently WIP -- we're still waiting on upstream odh-dashboard to add support for all the fields we care about.


samuelvl:
/hold

This PR will have `main` branch as target branch. We will merge it for RHODS 1.16 release.

deploy.sh
```diff
@@ -95,72 +101,62 @@ else
...
-    echo "ERROR: Jupyterhub RDS secret does not exist."
+## To be removed in 1.17 or greater
+kfnbc_migration=1
+oc get dc jupyterhub -n ${ODH_PROJECT} &> /dev/null || kfnbc_migration=0
+if [ "$kfnbc_migration" -eq 0 ]; then
+  echo "INFO: Fresh Installation, proceeding normally"
+else
+  echo "INFO: Migrating from JupyterHub to KFNBC, deleting old JupyterHub artifacts"
+
+  oc delete kfdef opendatahub
```
VaishnaviHire, line 112: Wouldn't we want to update `opendatahub` kfdef to just have `JupyterHub` resources to avoid deletion of other components?

anishasthana: Isn't that already the case?

VaishnaviHire: Yes. I missed that the PR is against `notebook-controller` branch

samuelvl, line 112: Will the postgres instance be automatically removed by doing this?

@VaishnaviHire @anishasthana

deploy.sh
```diff
@@ -95,72 +101,62 @@ else
...
+  DEFAULT_PVC_SIZE=$(oc get cm -n ${ODH_PROJECT} jupyterhub-cfg -o jsonpath="{.data.singleuser_pvc_size}")
+
+  oc get cm -n ${ODH_PROJECT} odh-jupyterhub-global-profile -o jsonpath="{.data.jupyterhub-singleuser-profiles\.yaml}" > tmp.yaml
+
+  sed -i "s/profiles/notebookSizes/g" tmp.yaml
+  sed -i "s/name: globals/name: Default/g" tmp.yaml
+  sed -i "s/^/  /" tmp.yaml
+  sed -i '1s/^/spec:\n/' tmp.yaml
+  sed -i 's/^\(\s*\)cpu:\s\([0-9]\+\)/\1cpu: \"\2\"/' tmp.yaml
+  yq -i eval-all '. as $item ireduce ({}; . *+ $item)' odh-dashboard/configs/odh-dashboard-config.yaml tmp.yaml
```
samuelvl, line 126: Is `yq` available in the deployer image? In CPaaS Dockerfile we cannot download it from github.

samuelvl (commented):
@anishasthana Can you change target branch to `main` instead of `notebook-controller`?

deploy.sh
VaishnaviHire: As discussed, given that we are removing the Postgres deletion code from operator, this will not delete the instance automatically. We would need to delete the instance manually.

anishasthana: It should right? The operator specific deletion code was only for uninstalling RHODS altogether. In this case the postgres object will get deleted normally.

anishasthana:
Replaced by #250

PR #255, opened 2022-08-10
Corrected kfnbc alert rule expressions to firing when the pods are unreachable

atheo89:
Corrected kfnbc alert rule expressions to firing when the pods are unreachable

## Description
While I was working on Prometheus alerts, I observed that the `RHODS Notebook controllers` alerts presented a false firing. 
The pods of the kfnbc and odh controllers on the other hand was working properly. 
The reason that the Firing alarm was triggered was because of missing in the end of the expression `== 0`

[Reference about how to alert for any instance that is unreachable](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#templating)

## How Has This Been Tested?
Ensure that the `notebook-controller-deployment-xxxxx-xxx` is down for more than 5 mins
Ensure that the `odh-notebook-controller-manager-xxxxx-xxx` is down for more than 5 mins

## Merge criteria:

- [ ] JIRA link(s):
- [ ] Live build image: 
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/monitoring/prometheus/prometheus-configs.yaml
+++ b/monitoring/prometheus/prometheus-configs.yaml
@@ -436,7 +436,7 @@ data:
             message: 'Kubeflow Notebook controller is down!'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/Jupyter/rhods-kfnbc-notebook-controller-alert.md"
             summary: Kubeflow notebook controller pod is not running
-          expr: sum(up{job=~'Kubeflow Notebook Controller Service Metrics'})
+          expr: absent(up{job=~'Kubeflow Notebook Controller Service Metrics'})
           for: 5m
           labels:
             severity: warning
@@ -445,7 +445,7 @@ data:
             message: 'ODH notebook controller is down!'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/Jupyter/rhods-odh-notebook-controller-alert.md"
             summary: ODH notebook controller pod is not running
-          expr: sum(up{job=~'ODH Notebook Controller Service Metrics'})
+          expr: absent(up{job=~'ODH Notebook Controller Service Metrics'})
           for: 5m
           labels:
             severity: warning


atheo89:
Adding in /cc @Samu to be aware for the PRs that deal with the changes on the monitoring stack

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -436,7 +436,7 @@ data:
             message: 'Kubeflow Notebook controller is down!'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/Jupyter/rhods-kfnbc-notebook-controller-alert.md"
             summary: Kubeflow notebook controller pod is not running
-          expr: sum(up{job=~'Kubeflow Notebook Controller Service Metrics'})
+          expr: up{job=~'Kubeflow Notebook Controller Service Metrics'} == 0
```
anishasthana, line 439: Will we only have one Notebook Controller pod?

samuelvl: Yes, same as `rhods-operator` controller pod. HA is delegated to Openshift API.

jgarciao: @atheo89 I've tested this alerts modifying manually the prometheus configmap and restarting the prometheus pod and I think it's not working well

Steps:
1. Scale down the operator deployment to zero pods
2. Scale down the odh-notebook-controller-manager deployment to zero pods
3. Wait more than 5 mins. The alert "ODH notebook controller pod is not running " is not fired

I believe the problem is that, when we scale down  odh-notebook-controller-manager deployment to zero pods, the expression `up{job=~'Kubeflow Notebook Controller Service Metrics'}` returns empty instead of 0 and the alert is not fired

accorvin: @atheo89 I think we've handled this in other places in the past by using an expression like this:
```
absent(up{job=~'Kubeflow Notebook Controller Service Metrics'})
```

atheo89: Great catch @jgarciao, I fixed it using the suggestion of @accorvin! and I tested as well, I didn't know about absent function, I was trying to solve it with `(sum(up{job=~'Kubeflow Notebook Controller Service Metrics'}  or on() vector(0))) == 0`

atheo89:
The changes of this PR have been merged into https://github.com/red-hat-data-services/odh-deployer/pull/256

PR #256, opened 2022-08-10
Update rhods prometheus stack according the kfnbc migration

atheo89:
This PR cleans up the obsolete monitoring targets, alerts, and recording rules from the Prometheus configuration

## Description
Applied the following changes:

- Removed Traefik
- Removed Jupyterhub and JupyterhubDB
- Added monitoring target for Jupyter notebook spawner page (Dependent by this PR https://github.com/red-hat-data-services/odh-deployer/pull/258)
- Updated the rhods_aggregate_availability metric to check also the spawner
- Corrected kfnbc alert rule expressions to firing when the pods are unreachable 

## How Has This Been Tested?

- Ensure that Traefik proxy doesn't exist on **recording rules** and on RHODS Probe Success Burn Rate **alert rules**
- Ensure that Jupyterhub doesn't exist on **recording rules** 
- To ensure that Usage Metrics work 
  - for **rhods_total_users** try to create and delete multiple notebooks with different users
  - for **rhods_active_users** try to create multiple notebooks with different users 
- Ensure that alerts for Jupyter notebook spawner are working
  - to check if the spawner reacts as expected delete the ServiceAccounts notebook-controller-service-account, odh-notebook-controller-manager, and rhods-dashboard and monitor the downtime via rhods_aggregate_availability (That is going to disturb the communication)
- Ensure that the rhods_aggregate_availability metric displays correctly the downtime of rhods-dashboard, notebook-spawner and combined
- To test the KFNBC alerts:
  - Ensure that the notebook-controller-deployment-xxxxx-xxx is down for more than 5 mins
  - Ensure that the odh-notebook-controller-manager-xxxxx-xxx is down for more than 5 mins

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-4765
- [x] Live build image: quay.io/accorvin/rhods-operator-live-catalog:1.0.0-[rhods-4765](https://issues.redhat.com//browse/rhods-4765)
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -217,11 +217,13 @@ pagerduty_service_token=$(echo -ne "$pagerduty_service_token" | tr -d "'" | base
 oc apply -f monitoring/jupyterhub-route.yaml -n $ODH_PROJECT
 oc apply -f monitoring/rhods-dashboard-route.yaml -n $ODH_PROJECT
 
-jupyterhub_host=$(oc::wait::object::availability "oc get route jupyterhub -n $ODH_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
 rhods_dashboard_host=$(oc::wait::object::availability "oc get route rhods-dashboard -n $ODH_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
 
-sed -i "s/<jupyterhub_host>/$jupyterhub_host/g" monitoring/prometheus/prometheus-configs.yaml
+NOTEBOOK_SUFFIX="\/notebookController\/spawner"
+notebook_spawner_host=$(oc::wait::object::availability "oc get route rhods-dashboard -n $ODH_PROJECT -o jsonpath='{.spec.host}'$NOTEBOOK_SUFFIX'" 2 30 | tr -d "'")
+
 sed -i "s/<rhods_dashboard_host>/$rhods_dashboard_host/g" monitoring/prometheus/prometheus-configs.yaml
+sed -i "s/<notebook_spawner_host>/$notebook_spawner_host/g" monitoring/prometheus/prometheus-configs.yaml
 sed -i "s/<pagerduty_token>/$pagerduty_service_token/g" monitoring/prometheus/prometheus-configs.yaml
 sed -i "s/<set_alertmanager_host>/$alertmanager_host/g" monitoring/prometheus/prometheus.yaml
 
@@ -278,9 +280,11 @@ alertmanager_config=$(oc get cm alertmanager -n $ODH_MONITORING_PROJECT -o jsonp
 
 sed -i "s#<prometheus_config_hash>#$prometheus_config#g" monitoring/prometheus/prometheus.yaml
 sed -i "s#<alertmanager_config_hash>#$alertmanager_config#g" monitoring/prometheus/prometheus.yaml
+sed -i "s#<odh_monitoring_project>#$ODH_MONITORING_PROJECT#g" monitoring/prometheus/prometheus-viewer-rolebinding.yaml
 
 oc apply -n $ODH_MONITORING_PROJECT -f monitoring/prometheus/prometheus.yaml
 oc apply -n $ODH_MONITORING_PROJECT -f monitoring/grafana/grafana-sa.yaml
+oc apply -n $ODH_PROJECT -f monitoring/prometheus/prometheus-viewer-rolebinding.yaml
 
 
 prometheus_route=$(oc::wait::object::availability "oc get route prometheus -n $ODH_MONITORING_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
--- a/monitoring/prometheus/prometheus-configs.yaml
+++ b/monitoring/prometheus/prometheus-configs.yaml
@@ -15,90 +15,6 @@ metadata:
 data:
   recording.rules: |
     groups:
-      - name: SLOs - JupyterHub
-        rules:
-        # The haproxy errors SLO for JupyterHub is disabled. Context at https://issues.redhat.com/browse/RHODS-2101
-
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[1d])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate1d
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[1h])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate1h
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[2h])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate2h
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[30m])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate30m
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[3d])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate3d
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[5m])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate5m
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"jupyterhub-.*", job="user_facing_endpoints_status"}[6h])
-          labels:
-            instance: jupyterhub
-          record: probe_success:burnrate6h
-
-      - name: SLOs - Traefik Proxy
-        rules:
-        - expr: min(up{job="Traefik Proxy Metrics"})
-          labels:
-            instance: traefik
-            job: user_facing_endpoints_status
-          record:
-            probe_success
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[1d])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate1d
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[1h])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate1h
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[2h])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate2h
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[30m])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate30m
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[3d])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate3d
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[5m])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate5m
-        - expr: |
-            1 - avg_over_time(probe_success{instance=~"traefik", job="user_facing_endpoints_status"}[6h])
-          labels:
-            instance: traefik
-          record: probe_success:burnrate6h
-
       - name: SLOs - ODH Dashboard
         rules:
         - expr: |
@@ -198,14 +114,18 @@ data:
 
       - name: Usage Metrics
         rules:
-        - expr: max(jupyterhub_total_users)
-          labels:
-            instance: jupyterhub
+        - expr: count(kube_statefulset_replicas{namespace=~"rhods-notebooks", statefulset=~"jupyter-nb-.*"})
           record: rhods_total_users
+          labels:
+            instance: jupyter-notebooks
+        - expr: count(kube_statefulset_replicas{namespace=~"rhods-notebooks", statefulset=~"jupyter-nb-.*"} ==1)
+          record: rhods_active_users
+          labels:
+            instance: jupyter-notebooks
 
       - name: Availability Metrics
         rules:
-        - expr: ((min(probe_success{instance=~"jupyterhub-redhat-ods-applications.*|rhods-dashboard-redhat-ods-applications.*"}) by (instance) or on() vector(0)) or label_replace(min(probe_success{instance=~"jupyterhub-redhat-ods-applications.*|rhods-dashboard-redhat-ods-applications.*"}), "instance", "combined", "instance", ".*"))
+        - expr: ((min(probe_success{name=~"rhods-dashboard|notebook-spawner"}) by (name) or on() vector(0)) or label_replace(min(probe_success{name=~"rhods-dashboard|notebook-spawner"}), "name", "combined", "name", ".*"))
           record: rhods_aggregate_availability
 
   alerting.rules: |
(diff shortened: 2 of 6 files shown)


atheo89:
Adding in /cc @Samu to be aware for the PRs that deal with the changes on the monitoring stack

atheo89:
Some notes and thoughts regarding the third point of this PR `Add monitoring target for Jupyter notebook spawner page`

**Case 1:**
 Adding the URLs 
`- rhods-dashboard-redhat-ods-applications.apps.atheodor.dev.datahub.redhat.com/notebookController/spawner` or 
`- rhods-dashboard-redhat-ods-applications.apps.atheodor.dev.datahub.redhat.com/notebookController`
as targets into the job: `user_facing_endpoints_status` and alongside deleting the ServiceAccounts `notebook-controller-service-account`, `odh-notebook-controller-manager` and `rhods-dashboard` (Having the operator scaled down) I don't see any downtime on the targets via `rhods_aggregate_availability` metric. (The scraping of the targets is UP)

**Case 2:**
Adding the URLs 
`- rhods-dashboard-redhat-ods-applications.apps.atheodor.dev.datahub.redhat.com/notebookController/spawner` or 
`- rhods-dashboard-redhat-ods-applications.apps.atheodor.dev.datahub.redhat.com/notebookController`
as targets into the job: `user_facing_endpoints_status` and stop notebook server (via rhods dashboard) I don't see any downtime again through `rhods_aggregate_availability` metric.

I believe the reason that we don't see any downtime while we scrape these targets is that the kfnbc spawner is embedded on rhods dashboard.

Any alternative thoughts regarding how to scrape or not the notebook spawner?

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -196,16 +112,9 @@ data:
             instance: rhods-controller
           record: controller_runtime_reconcile_total:rate15m
 
-      - name: Usage Metrics
-        rules:
-        - expr: max(jupyterhub_total_users)
-          labels:
-            instance: jupyterhub
-          record: rhods_total_users
```
accorvin, line 204: I know that the BU watches this metric pretty closely to understand how much notebooks are being used, so we should try to retain this measurement if possible. Do you know if KFNBC has any built in metrics that track the number of users?

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -298,63 +207,15 @@ data:
...
-            triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
-            summary: RHODS JupyterHub Probe Success Burn Rate
-          expr: |
-            sum(probe_success:burnrate6h{instance=~"jupyterhub|jupyterhub-db"}) by (instance) > (1.00 * (1-0.98000))
-            and
-            sum(probe_success:burnrate3d{instance=~"jupyterhub|jupyterhub-db"}) by (instance) > (1.00 * (1-0.98000))
-          for: 3h
-          labels:
-            severity: warning
         - alert: RHODS Probe Success Burn Rate
```
accorvin, line 210: MT-SRE has requested that we split this alert up so that the alert name is more specific to the service that is alerting. In its current state, this alert will fire for failures on both the RHODS dashboard and the notebook spawner.

I think we need to copy this alert block and then change the expressions so that one fires only for the dashboard and the other fires only for the notebook spawner.

monitoring/prometheus/prometheus-configs.yaml
atheo89: Maybe this one could work `runningNotebooks` https://github.com/opendatahub-io/kubeflow/blob/321ead60e215a3eb9d579a1e3de657150606e24d/components/notebook-controller/pkg/metrics/metrics.go#L15

accorvin: I fear that that's the number of actively running users (which is still a very useful metric to have - we have some metrics around this already which probably need to be updated). To get the total number of all time users we may need to generate it from something else, perhaps the number of PVCs matching the right name pattern or something like that.

monitoring/prometheus/prometheus-configs.yaml
atheo89: Added the monitoring target for Jupyter notebook spawner page, and I updated the alert rule to fire the notebook spawner. 
relevant commits:

- [Added monitoring target for Jupyter notebook spawner page](https://github.com/red-hat-data-services/odh-deployer/pull/256/commits/03e964a863a47ecfaa422400ca1ef8935ef1ab3a)
- [Added alert block that fires the notebook spawner](https://github.com/red-hat-data-services/odh-deployer/pull/256/commits/d6c59a9e92058df1d444d6ebd086531f42412682)

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -195,17 +111,21 @@ data:
           labels:
             instance: rhods-controller
           record: controller_runtime_reconcile_total:rate15m
-
+
```
accorvin, line 114: nit: looks like you're adding trailing whitespace here. Is that deliberate?

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -364,9 +284,9 @@ data:
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
             summary: RHODS Probe Success Burn Rate
           expr: |
-            sum(probe_success:burnrate30m{instance=~"rhods-dashboard|traefik"}) by (instance) > (6.00 * (1-0.99950))
+            sum(probe_success:burnrate30m{name=~"rhods-dashboard"}) by (name) > (6.00 * (1-0.99950))
```
accorvin, line 287: For all of the alerting rules where you're making this change, the alerts are named `RHODS Probe Success Burn Rate`. Since these are now specific to the dashboard, can you change the name to `RHODS Dashboard Probe Success Burn Rate`? 

This change:
1. Makes these alert names consistent with the changes you've made above for the jupyter alerts
2. Satisfies MT-SRE's request to make the alert names more descriptive.

monitoring/prometheus/prometheus-configs.yaml
atheo89: Yep, it separates visually the ` SLOs - RHODS Operator` rules from  `Usage Metrics`

monitoring/prometheus/prometheus-configs.yaml
atheo89: Thanks for noticing that! Changed.

monitoring/prometheus/prometheus-configs.yaml
accorvin: I'm not talking about the empty line, but rather the fact that the new line includes 4 spaces on it. (good yaml syntax dictates that blank lines contain no whitespace characters e.g. spaces or tabs)

accorvin:
@atheo89 these changes look good to me (other than the small whitespace nit). Do you have a live build I can test with?

monitoring/prometheus/prometheus-configs.yaml
atheo89: Ahhh ok, the spaces added by vs code. I deleted them and kept only the newline

atheo89:
@accorvin today, we had a discussion with @samuelvl about the live build. He is going to bunch all the monitoring changes together and he will provide a live build. Once it will be ready we will add the link for the live build.

anishasthana:
@atheo89 the prometheus stack init container will also need to be updated to account for traefik proxy
https://github.com/red-hat-data-services/odh-deployer/blob/main/monitoring/prometheus/prometheus.yaml#L22-L35

atheo89:
@anishasthana, I should remove the command, right?

anishasthana:
It would be better to update it so that prometheus is waiting for one of the notebook controller pods to come up.

atheo89:
**Note**: Added on this PR the changes for the fault firing of the kfnbc alerts, and I closed the old one.  https://github.com/red-hat-data-services/odh-deployer/pull/255

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -298,111 +218,111 @@ data:
 
       - name: SLOs-probe_success
         rules:
-        - alert: RHODS JupyterHub Probe Success Burn Rate
+        - alert: RHODS Jupyter Probe Success Burn Rate
           annotations:
             message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
```
jgarciao, line 224: triage url should be updated for kubeflow

jgarciao (changes requested):

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -298,111 +218,111 @@ data:
...
-            sum(probe_success:burnrate1d{instance=~"jupyterhub|jupyterhub-db"}) by (instance) > (3.00 * (1-0.98000))
+            sum(probe_success:burnrate1d{name=~"notebook-spawner"}) by (name) > (3.00 * (1-0.98000))
           for: 1h
           labels:
             severity: warning
-        - alert: RHODS JupyterHub Probe Success Burn Rate
+        - alert: RHODS Jupyter Probe Success Burn Rate
           annotations:
             message: 'High error budget burn for {{ $labels.instance }} (current value: {{ $value }}).'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/JupyterHub/rhods-probe-success-burn-rate.md"
```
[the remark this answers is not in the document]
atheo89: Updated

accorvin:
A new live build with these changes is available at quay.io/accorvin/rhods-operator-live-catalog:1.0.0-rhods-4765

accorvin:
@atheo89 I'm getting an error in the deployer init container:
```
sed: -e expression #1, char 101: unknown option to `s'
```

deploy.sh
```diff
@@ -217,11 +217,13 @@ pagerduty_service_token=$(echo -ne "$pagerduty_service_token" | tr -d "'" | base
 oc apply -f monitoring/jupyterhub-route.yaml -n $ODH_PROJECT
 oc apply -f monitoring/rhods-dashboard-route.yaml -n $ODH_PROJECT
 
-jupyterhub_host=$(oc::wait::object::availability "oc get route jupyterhub -n $ODH_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
 rhods_dashboard_host=$(oc::wait::object::availability "oc get route rhods-dashboard -n $ODH_PROJECT -o jsonpath='{.spec.host}'" 2 30 | tr -d "'")
 
-sed -i "s/<jupyterhub_host>/$jupyterhub_host/g" monitoring/prometheus/prometheus-configs.yaml
+NOTEBOOK_SUFFIX="/notebookController/spawner"
+notebook_spawner_host=$(oc::wait::object::availability "oc get route rhods-dashboard -n $ODH_PROJECT -o jsonpath='{.spec.host}'$NOTEBOOK_SUFFIX'" 2 30 | tr -d "'")
```
accorvin, line 223: Is there an extra `'` character here right after `$NOTEBOOK_SUFFIX` ?

accorvin: Nope, this was not the issue.

accorvin:
I believe I've found the problem that was causing the deploy script to fail. I've added https://github.com/red-hat-data-services/odh-deployer/pull/256/commits/b9eb01445121d4315bc2833829d65ae742d37be5 with a fix and am building a new live image now.

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -436,7 +356,7 @@ data:
             message: 'Kubeflow Notebook controller is down!'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/Jupyter/rhods-kfnbc-notebook-controller-alert.md"
             summary: Kubeflow notebook controller pod is not running
-          expr: sum(up{job=~'Kubeflow Notebook Controller Service Metrics'})
+          expr: absent(up{job=~'Kubeflow Notebook Controller Service Metrics'})
           for: 5m
           labels:
             severity: warning
```
jgarciao, line 362: If severity = warning SRe won't be paged if notebook controller is down for more than  5 minutes. Should this be critical instead?

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -445,7 +365,7 @@ data:
             message: 'ODH notebook controller is down!'
             triage: "https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/RHODS/Jupyter/rhods-odh-notebook-controller-alert.md"
             summary: ODH notebook controller pod is not running
-          expr: sum(up{job=~'ODH Notebook Controller Service Metrics'})
+          expr: absent(up{job=~'ODH Notebook Controller Service Metrics'})
           for: 5m
           labels:
             severity: warning
```
jgarciao, line 371: If severity = warning SRe won't be paged if notebook controller is down for more than  5 minutes. Should this be critical instead?

deploy.sh
jgarciao, line 223: This variable contains the route to the new spawner page (e.g. https://rhods-dashboard-redhat-ods-applications.apps.qeaisrhods-mon.XXXX.s1.devshift.org/notebookController/spawner), that is used in the alert _RHODS Jupyter Probe Success Burn Rate_

What I have experimented in the livebuild is that, if the _Kubeflow notebook controller_ pod and the _ODH notebook controller_ pod are not running, the alert _RHODS Jupyter Probe Success Burn Rate_ does not fire.

I believe the reason is that https://rhods-dashboard-redhat-ods-applications.apps.qeaisrhods-mon.XXXX.s1.devshift.org/notebookController/spawner renders the page correctly (returning HTTP 200) even if those pods are not running

What happens in that situation is that, when you try to spawn a notebook, you get this error:
error-spawning-notebook-when-pods-are-not-running

So, I don't know if the alert _RHODS Jupyter Probe Success Burn Rate_ is useful with this implementation, having already  _RHODS Dashboard Probe Success Burn Rate_

jgarciao (changes requested):

deploy.sh
accorvin: @lucferbux @andrewballantyne this to me seems like functionality that should perhaps change on the notebook spawner page. If the backend dependencies aren't met, does it make sense for the spawner page to generate a 200 status code when you load it?

monitoring/prometheus/prometheus-configs.yaml
accorvin: When we met with SRE a few weeks ago to come up with the plan for monitoring KFNBC, we agreed that we would make this a warning level error while/until we determine how the alert behaves in production. The thought was that we should be able to rely on kubernetes functionality to ensure that the pod is always running, so we likely don't need any alert for this. We weren't 100% confident of this decision though so a warning alert seemed like a good compromise.

monitoring/prometheus/prometheus-configs.yaml
accorvin: When we met with SRE a few weeks ago to come up with the plan for monitoring KFNBC, we agreed that we would make this a warning level error while/until we determine how the alert behaves in production. The thought was that we should be able to rely on kubernetes functionality to ensure that the pod is always running, so we likely don't need any alert for this. We weren't 100% confident of this decision though so a warning alert seemed like a good compromise.

accorvin:
The issue with the odh deployer init container failing has been resolved by my latest change (this change is available in the live build, I pushed the new version to the same tag).

I'm seeing an issue now where the probes are getting 403 errors. I'm investigating why that is happening.

accorvin:
Update: I found that the opendatahub-operator container in my live build did not properly include the changes from https://github.com/red-hat-data-services/odh-manifests/pull/231

I'm generating a new live build that resolves that and will post with the status of my testing.

accorvin:
Update: this looks to be working correctly now with the latest version of my live build

deploy.sh
cfchase: I don't think we want the dashboard in the business of checking for status of backend controllers.  I think we'd do a poor job of it since we can't really anticipate the ways in which it fails for any given controller.  Honestly, I think this UX is pretty good the way it is since it's properly conveying an error from the notebook spawn.

LaVLaS:
Rhods livebuild for these changes with KFNBC and migration script: [quay.io/llasmith/rhods-operator-live-catalog:1.16.0-dashboard-kfnbc](http://quay.io/llasmith/rhods-operator-live-catalog:1.16.0-dashboard-kfnbc)

monitoring/prometheus/prometheus-configs.yaml
jgarciao: Ok, it makes sense. Thanks!

monitoring/prometheus/prometheus-configs.yaml
jgarciao: Ok, it makes sense. Thanks!

anishasthana:
We've pulled Adriana's changes into #257

deploy.sh
accorvin: We definitely need to change something around how this works - either what we’re probing or how the page behaves. Note: I don’t think we should make this change now, but note it as a deficiency and fix it in 1.17.

Right now, kfnbc functionality other than the spawner UI can be broken and we wouldn’t know about it.

cfchase: It should be easy to add liveness probes to all the backend controllers so we aren't ever in this situatoin, shouldn't it?

atheo89:
These changes have been pulled into https://github.com/red-hat-data-services/odh-deployer/pull/257

PR #257, opened 2022-08-15
Migrate completely from JupyterHub to KFNBC

anishasthana:
- Delete old JupyterHub artifacts as part of migration to kfnbc
- Move CUDA Buildchain to kfnbc directory, remove old 11.0.3 chain
- Migrate RHODS Group Configs
- Migrate RHODS Groups to ODH dashboard Config
- Migrate Jupyterhub default settings
- Migrate notebook sizes to nbc

This replaces #250 . The reason I'm opening this is due to the rebase on main being very complex.

cc @VaishnaviHire @samuelvl


deploy.sh
```diff
@@ -95,73 +101,62 @@ else
...
+  DEFAULT_PVC_SIZE=$(oc get cm -n ${ODH_PROJECT} jupyterhub-cfg -o jsonpath="{.data.singleuser_pvc_size}")
+
+  oc get cm -n ${ODH_PROJECT} odh-jupyterhub-global-profile -o jsonpath="{.data.jupyterhub-singleuser-profiles\.yaml}" > tmp.yaml
+
+  sed -i "s/profiles/notebookSizes/g" tmp.yaml
+  sed -i "s/name: globals/name: Default/g" tmp.yaml
+  sed -i "s/^/  /" tmp.yaml
+  sed -i '1s/^/spec:\n/' tmp.yaml
+  sed -i 's/^\(\s*\)cpu:\s\([0-9]\+\)/\1cpu: \"\2\"/' tmp.yaml
+  yq -i eval-all '. as $item ireduce ({}; . *+ $item)' odh-dashboard/configs/odh-dashboard-config.yaml tmp.yaml
```
samuelvl, line 126: Is `yq` installed in the deployer image?

Dockerfile
```diff
@@ -28,8 +28,7 @@ COPY buildchain.sh $HOME
 ADD kfdefs $HOME/kfdefs
 ADD monitoring $HOME/monitoring
 ADD consolelink $HOME/consolelink
-ADD groups $HOME/groups
-ADD jupyterhub $HOME/jupyterhub
+ADD kfnbc $HOME/kfnbc
```
samuelvl, line 31: I'd use `notebook-controller` or `nbc` instead of `kfnbc` to avoid using kubeflow references in downstream.

deploy.sh
```diff
@@ -95,73 +101,62 @@ else
...
+  oc delete -n ${ODH_PROJECT} secret jupyterhub-database-secret
+  oc delete -n ${ODH_PROJECT} configmap jupyterhub-cfg
+  oc delete -n ${ODH_PROJECT} configmap odh-jupyterhub-global-profile
+  oc delete -n ${ODH_PROJECT} configmap rhods-jupyterhub-sizes
+
+  oc delete -n ${ODH_PROJECT} -f cloud-resource-operator/crds
+  oc delete -n ${ODH_PROJECT} -f cloud-resource-operator/rbac
+  oc delete -n ${CRO_PROJECT} -f cloud-resource-operator/rbac-rds
+  oc delete -n ${CRO_PROJECT} -f cloud-resource-operator/deployment
+  oc delete -n ${ODH_PROJECT} -f cloud-resource-operator/postgres.yaml
```
samuelvl, line 141: Should we wait here until the postgres database is removed?

samuelvl (commented):
Could you provide a live build image to test the migration?

deploy.sh
```diff
@@ -393,6 +351,13 @@ exists=$(oc get -n $ODH_PROJECT ${kind} ${object} -o name | grep ${object} || ec
 #TODO: This controls feature flags and notebook controller presets like Notebook size. Confirm that notebook sizes can be configured external to the ODHDashboardConfig CR
```
cfchase, line 351: Won't odh-dashboard-confg always exist from the 1.15 installation but the `notebookController.enabled` would be blank or false?  If that's the case, it seems like you're always going to skip migration?

odh-dashboard/configs/odh-dashboard-config.yaml
```diff
@@ -13,7 +13,13 @@ spec:
...
     enablement: true
+  groupsConfig:
+    admin_groups: "<admin_groups>"
+    allowed_groups: "<allowed_groups>"
   notebookController:
+    cullerTimeout: <timeout>
+    defaultPersistentVolumeClaimSize: <size>
+    enabled: true
     notebookNamespace: rhods-notebooks
   notebookSizes:
```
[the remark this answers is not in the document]
samuelvl: After scrum meeting we decided to create the Default size in the deployer:

```yaml
  notebookSizes:
    - name: Default
      resources:
        limits:
          cpu: "2"
          memory: 8Gi
        requests:
          cpu: "1"
          memory: 4Gi
...
```

cc @DaoDaoNoCode

deploy.sh
VaishnaviHire: The cloud resource operator is responsible for handling clean deletion of postgres instance and any dependent resources.

We should delete the `postgres` instance, ensure its deletion is completed before deleting the operator pod(cloud-resource-operator/deployment).

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -485,45 +485,6 @@ data:
...
-      relabel_configs:
-        - source_labels: [__meta_kubernetes_service_name]
-          regex: ^(traefik-proxy)$
-          target_label: kubernetes_name
-          action: keep
-        - source_labels: [__address__]
-          regex: (.+):(\d+)
-          target_label: __address__
-          replacement: ${1}:8082
-
```
VaishnaviHire, line 526: Do we also need to remove JupyterHub alerts (e.g RHODS JupyterHub Probe Success Burn Rate) ?

deploy.sh
anishasthana: yq is not in the default deployer image -- I figured we could install it as part of the image generation.

odh-dashboard/configs/odh-dashboard-config.yaml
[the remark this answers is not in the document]
cfchase: I think this would be ok, but can you guys review this with the size configuration mockups?  I seem to remember the mockups for default being a checkbox selecting an existing size.   You could set default as it's own size or add a tag to an existing size, whatever makes more sense.

monitoring/prometheus/prometheus-configs.yaml
anishasthana: Adriana worked on that as part of #256.

odh-dashboard/configs/odh-dashboard-config.yaml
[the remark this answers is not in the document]
anishasthana: It was eventually decided that we should _not_ have a default profile. https://github.com/opendatahub-io/odh-dashboard/issues/345#issuecomment-1217660032 has the full context.

deploy.sh
anishasthana: Good catch.

deploy.sh
```diff
@@ -95,73 +101,80 @@ else
...
+  while [ oc oc get postgres -n {ODH_PROJECT} jupyterhub-db-rds -o yaml ]
+  do
+    echo "Waiting for postgres object to be deleted"
+    ((jj=jj+1))
+    # Wait for 30 minutes
+    if [ $jj -eq 360 ]; then
+        echo "Postgress deletion failed"
+        exit 1
     fi
+    sleep 5
```
cfchase, line 153: Are we going to have 20 minutes of downtime waiting for this delete?  

If so we're either going to have to wait and delete the cro in the background, **or** do something else clever (like a cleanup job executed later), **or** we might leave the CRO deployment for another release.

anishasthana:
Corresponding Live build PR: https://gitlab.cee.redhat.com/data-hub/rhods-live-builder/-/merge_requests/30

deploy.sh
anishasthana: That's actually what I wanted to do initially -- I figured we just delete the postgres object this time and then delete the CRO in 1.17. I'm more than happy to change it to that -- it'll greatly simplify the uninstall too.

Dockerfile
```diff
@@ -23,13 +23,15 @@ RUN tar -C /usr/local/bin -xvf $TMPDIR/oc.tar.gz && \
...
 COPY buildchain.sh $HOME
 ADD kfdefs $HOME/kfdefs
 ADD monitoring $HOME/monitoring
 ADD consolelink $HOME/consolelink
-ADD groups $HOME/groups
-ADD jupyterhub $HOME/jupyterhub
+ADD nbc $HOME/nbc
 ADD partners $HOME/partners
 ADD network $HOME/network
 ADD cloud-resource-operator $HOME/cloud-resource-operator
```
anishasthana, line 37: On second thought... we could remove this altogether as the files aren't required. We can simply update the oc commands in deployer to refer directly to the object names, no yaml needed. Thoughts?

anishasthana (commented):
Live build is up at quay.io/anishasthana/rhods-operator-live-catalog:1.16.1-4600.
I've done a couple migration tests and fresh installs and not seen any major issues.... yet. I'll be testing further tomorrow, but I feel comfortable sharing it out now to the wider team for review

One open "question" for me is https://github.com/red-hat-data-services/opendatahub-operator/tree/rhods_4600. I did a best-effort update of the odh-dashboard manifests in here but it's possible I made a mistake. @samuelvl / @maroroman I'd appreciate eyes on that.

Dockerfile
samuelvl: I'm fine with that

samuelvl:
@pablofelix @tarukumar You are the QE approvers in this repository but I have been working more closely with @lugi0 during the notebook controller development, let me know who is our contact to start testing the migration script. We want to ensure that users do not lose data and the migration is successful:

1. Install RHODS 1.15
2. Create several notebooks (user1, user2, etc) with the following information:
    - Plain (configmap) environment variable.
    - Secret environment variable.
    - Add data to PVC.
3. Update to RHODS 1.16 (using the live build)
4. Spawn user notebooks with the same configuration.
5. Verify data is still accessible.

After the migration is completed, we should also check that all the requirements in https://github.com/opendatahub-io/odh-dashboard/issues/213 are covered and working.

samuelvl:
@anishasthana Changes in https://github.com/red-hat-data-services/odh-manifests/tree/rhods_4600 look fine, only one comment. I wouldn't add `odh-dashboard/overlays/odhdashboardconfig/*` files because in RHODS this configuration is stored in the deployer.

anishasthana:
I'd generated quay.io/anishasthana/rhods-operator-live-catalog:1.16.2-4600 on Friday but the conference prevented me from fully testing it out. Something seems wrong with the notebook size migration -- I'll be looking into it on Monday.

tarukumar:
> 

@samuelvl @pablofelix and me will be working on this feature. Let us know when the live build will be available so that we can create environment for the validation

samuelvl:
@anishasthana I tried upgrading from 1.15 and the opendatahub kfdef is not getting removed (hence jh stack either) after more than 1 hour waiting:

```
$ oc get kfdef opendatahub -o yaml
apiVersion: kfdef.apps.kubeflow.org/v1
kind: KfDef
metadata:
  annotations:
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2022-08-23T10:38:39Z"
  finalizers:
  - kfdef-finalizer.kfdef.apps.kubeflow.org
```

In the deployer logs I can see the `oc delete` succeed but we may need forcing the deletion: 

```
$ oc logs -l name=rhods-operator -c rhods-deployer --tail=-1 -n redhat-ods-operator
...
INFO: Migrating from JupyterHub to NBC, deleting old JupyterHub artifacts
kfdef.kfdef.apps.kubeflow.org "opendatahub" deleted
```

anishasthana:
```
Latest live build is up at quay.io/anishasthana/rhods-operator-live-catalog:1.16.3-4600
```
@tarukumar we've had (broadly working) live builds up. The ones that are being generated are to address small bugs/functionality improvements being found/suggested by dev. I think we should get more eyes on it faster so that the bugs are identified right now instead of later.

anishasthana:
I built a new catalogsource which included Vaishnavi's operator changes[1]: quay.io/anishasthana/rhods-operator-live-catalog:1.16.4-4600

[1]https://github.com/red-hat-data-services/opendatahub-operator/pull/152

tarukumar:
> 

@anishasthana we will stat looking into it. I am planning to create cluster with rhods version `quay.io/modh/qe-catalog-source:v1150-10` to perform upgrade with above live build. Let me know if that sounds good

anishasthana:
That's the one I've been using for testing too!

pablofelix:
@tarukumar and me, were testing the upgrade and we found some issues in the migration:

- There is a downtime of 10 minutes in Jupyterhub and the Jupyterhub pods are running after those 10 mins instead of being deleted. 
Screenshot from 2022-08-24 14-49-44

- Spawning some notebooks we found two different problems.
    - In the standard notebook, we found an error message on the first try. On the second try, it works.
Screenshot from 2022-08-24 12-54-42

    - In the minimal python notebook image we had to increase the node numbers to make it work because we had an error about insufficient resources. Before the upgrade was working with two workers nodes and two infra nodes. We had to add a worker node
    
- We also found  in the description of installed operators and in the notebook secrets names that we are still using 'jupyterhub' name

anishasthana:
(I've reached out offline as well)

> There is a downtime of 10 minutes in Jupyterhub and the Jupyterhub pods are running after those 10 mins instead of being deleted.
JupyterHub is taking some time to be deleted as we rely on the operator clean it up correctly. I've seen deletion take up to 20 minutes or more, depending on how the operator is feeling. 

> Spawning some notebooks we found two different problems.
That's odd, I didn't run into any issues with spawning during any of my tests.

> We also found in the description of installed operators and in the notebook secrets names that we are still using 'jupyterhub' name
Can you tell me the specific secrets? I'm assuming under `installed operators` will just require us to update the contents of managed tenant bundles.

tarukumar:
We also observed after the upgrade traefik pods are not deleted .

Screenshot from 2022-08-24 18-56-26

anishasthana:
Latest image is up at http://quay.io/anishasthana/rhods-operator-live-catalog:1.16.5-4600
This is how things look after a migration:
image
I was able to spin up a notebook pod without issue as well.

deploy.sh
cfchase: This code `"$exists" == "false"` is weirding me out.
```
#If the odh-dashboard-config resource does NOT exist, 
#  if the (nonexistent) resource is safe
#    apply changes to a nonexistent resource
#  else (the nonexistent resource is modified by whom I don't know since I've not seen this label actually used)
#     skip applying changes
#  end
#else
#  migrate on an existing odh-dashboard-config

if [ "$exists" == "false" ]; then
  if oc::object::safe::to::apply ${kind} ${resource}; then
    oc apply -n ${ODH_PROJECT} -f odh-dashboard/configs/odh-dashboard-config.yaml
  else
    echo "The ODHDashboardConfig (${kind}/${resource}) has been modified. Skipping apply."
  fi
else # Migration code for 1.16
```

Does `"$exists" == "false"` mean something else? 

I just want to make sure under any circumstance, we create an odh-dashboard-config with `notebookController.enabled = true` since the upstream default is still JupyterHub

tarukumar:
@pablofelix and me tested the upgrade with the new build provided. Please find our observation below:

1. We observed that while changing the PVC size from RHODS dashboard. Error msg is observed. This only occur when your 'Stop idle notebooks' set to 'Do not stop idle notebooks'.Even with the error when we try to spawn a new notebook it create PVC size which was set. From what we understand it's an UI issue.
Screenshot from 2022-08-25 17-22-24


2. Notebook secert and config map are still using the using 'jupyterhub' name. ex:jupyterhub-singleuser*
Screenshot from 2022-08-25 17-12-13


3. RHODS description still consists of jupyterhub under installed operators
Screenshot from 2022-08-25 17-18-28


4. Observed that ` rhods-groups-config` config map is removed.Do you know how could we change groups in kfnbc? (asked in slack as well)
Screenshot from 2022-08-25 17-25-17

5. We also observed after starting the notebook. We get popup for login again

6. When we click on 'Hub Control panel' we get 'Page not found' for some microsecond.We believe it is not able to find the directory to load image again this could be UI issue.

anishasthana:
> We observed that while changing the PVC size from RHODS dashboard. Error msg is observed. This only occur when your 'Stop idle notebooks' set to 'Do not stop idle notebooks'.Even with the error when we try to spawn a new notebook it create PVC size which was set. From what we understand it's an UI issue.

@andrewballantyne / @maroroman can you take a look at ^^ This may need a new issue


> We also observed after starting the notebook. We get popup for login again
> When we click on 'Hub Control panel' we get 'Page not found' for some microsecond.We believe it is not able to find the directory to load image again this could be UI issue.

@andrewballantyne fyi^^

> Notebook secert and config map are still using the using 'jupyterhub' name. ex:jupyterhub-singleuser*

This was done by design due to fears around complexity of migration. @DaoDaoNoCode  can confirm ^^

> RHODS description still consists of jupyterhub under installed operators

This will require an update in the managed tenants repo I believe. I'll see what to do here, but I'd like to not block the pR on this (as it is tied to either rhods-live-builder or CPaaS configurations). We will make sure to keep these changes in mind when the bundles are generated. @crobby fyi^^


> Observed that rhods-groups-config config map is removed.Do you know how could we change groups in kfnbc? (asked in slack as well)

Answered in slack. There is a CR field that can be updated.


> traefik-rules config map should get deleted during/after the upgrade.

Good catch. I'm adding the delete in now.

As you know alerts in Prometheus are still looking for JH . This changes will happen if we get build with this https://github.com/red-hat-data-services/odh-deployer/pull/256

> Do we need jupyterhub-idle-culler secret. If it is not need t we should delete it. In case we need this we need to rename it and remove JH from name

Good catch. I'm adding the delete in now.

andrewballantyne:
@anishasthana Thanks for the tags. I'll make sure we have something for them. Read below for my understanding/explantation. Do let me know if you see anything that is 1.16 critical.

---

> > We observed that while changing the PVC size from RHODS dashboard. Error msg is observed. This only occur when your 'Stop idle notebooks' set to 'Do not stop idle notebooks'.Even with the error when we try to spawn a new notebook it create PVC size which was set. From what we understand it's an UI issue.
> 
> @andrewballantyne / @maroroman can you take a look at ^^ This may need a new issue

Interesting -- this is definitely something I am unaware of. I'll make sure we have something logged upstream. Doesn't sound like a blocker though.
 
> > We also observed after starting the notebook. We get popup for login again
> > When we click on 'Hub Control panel' we get 'Page not found' for some microsecond.We believe it is not able to find the directory to load image again this could be UI issue.
> 
> @andrewballantyne fyi^^

Yeah, this was introduced in the "is allowed" PR that checked to see if you have auth based on a config (rather than just assuming if they can login they can create notebooks). This needs an issue, I have been meaning to do so. I'll make sure this is logged. It's a minor flicker because the state behind figuring out your permissions is async and takes a brief moment to update the UI -- after it has already loaded.

anishasthana:
image
I can confirm that the rhods groups config was migrated over.

anishasthana:
http://quay.io/llasmith/rhods-operator-live-catalog:1.16.0-dashboard-kfnbc consolidates a lot of the open PRs right now.

This contains monitoring fixes (so that monitoring stack comes up), alert updates, all the odh-dashboard updates that have been merged so far, all the work that went into migration, and Vaishnavi's operator update.

tarukumar:
Thanks @anishasthana . I observed one more issue with the http://quay.io/llasmith/rhods-operator-live-catalog:1.16.0-dashboard-kfnbc.

When saving new settings at Dashboard > Settings > Cluster Settings I get this error on the cluster with OSD 4.11. Do you know if it's being tracked already as an issue? I think this will be blocker since it is not allowing use to change any setting in Dashboard.
PS: Same is not Observed on OSD 4.10
OSD4.11:
Screenshot from 2022-08-26 14-06-28
OSD 4.10:
Screenshot from 2022-08-26 14-11-26

tarukumar:
> @anishasthana Thanks for the tags. I'll make sure we have something for them. Read below for my understanding/explantation. Do let me know if you see anything that is 1.16 critical.
> 
> > > We observed that while changing the PVC size from RHODS dashboard. Error msg is observed. This only occur when your 'Stop idle notebooks' set to 'Do not stop idle notebooks'.Even with the error when we try to spawn a new notebook it create PVC size which was set. From what we understand it's an UI issue.
> > 
> > 
> > @andrewballantyne / @maroroman can you take a look at ^^ This may need a new issue
> 
> Interesting -- this is definitely something I am unaware of. I'll make sure we have something logged upstream. Doesn't sound like a blocker though.
> 
> For posterity, logged [opendatahub-io/odh-dashboard#443](https://github.com/opendatahub-io/odh-dashboard/issues/443)
> 
> > > We also observed after starting the notebook. We get popup for login again
> > > When we click on 'Hub Control panel' we get 'Page not found' for some microsecond.We believe it is not able to find the directory to load image again this could be UI issue.
> > 
> > 
> > @andrewballantyne fyi^^
> 
> Yeah, this was introduced in the "is allowed" PR that checked to see if you have auth based on a config (rather than just assuming if they can login they can create notebooks). This needs an issue, I have been meaning to do so. I'll make sure this is logged. It's a minor flicker because the state behind figuring out your permissions is async and takes a brief moment to update the UI -- after it has already loaded.
> 
> For posterity, logged [opendatahub-io/odh-dashboard#444](https://github.com/opendatahub-io/odh-dashboard/issues/444)

@andrewballantyne I don't think this is blocker but it would be good if it's get fixed on 1.16. I think it would not give first good impression/experience to the customer after the migration.

andrewballantyne:
> @andrewballantyne I don't think this is blocker but it would be good if it's get fixed on 1.16. I think it would not give first good impression/experience to the customer after the migration.

@tarukumar While investigating an issue noted in a slack thread, I think I found out the solution to https://github.com/opendatahub-io/odh-dashboard/issues/443 -- also a workaround.

The workaround:
> 1) Toggle the tolerations checkbox on/off
> 2) Save nothing else

After that you should be good to go -- the issue is a lack of a default, definitely a bug but it seems to be an initial state problem.

tarukumar:
> > @andrewballantyne I don't think this is blocker but it would be good if it's get fixed on 1.16. I think it would not give first good impression/experience to the customer after the migration.
> 
> @tarukumar While investigating an issue noted in a slack thread, I think I found out the solution to [opendatahub-io/odh-dashboard#443](https://github.com/opendatahub-io/odh-dashboard/issues/443) -- also a workaround.
> 
> The workaround:
> 
> > 1. Toggle the tolerations checkbox on/off
> > 2. Save nothing else
> 
> After that you should be good to go -- the issue is a lack of a default, definitely a bug but it seems to be an initial state problem.



> > @andrewballantyne I don't think this is blocker but it would be good if it's get fixed on 1.16. I think it would not give first good impression/experience to the customer after the migration.
> 
> @tarukumar While investigating an issue noted in a slack thread, I think I found out the solution to [opendatahub-io/odh-dashboard#443](https://github.com/opendatahub-io/odh-dashboard/issues/443) -- also a workaround.
> 
> The workaround:
> 
> > 1. Toggle the tolerations checkbox on/off
> > 2. Save nothing else
> 
> After that you should be good to go -- the issue is a lack of a default, definitely a bug but it seems to be an initial state problem.

This issue is different from what is reported in slack. slack issue is mentioned above. https://github.com/opendatahub-io/odh-dashboard/issues/443 issue was observed when there was no  tolerations checkbox present on the dashboard. you can verify that in the images attached in this PR in original comment

andrewballantyne:
> This issue is different from what is reported in slack. slack issue is mentioned above. https://github.com/opendatahub-io/odh-dashboard/issues/443 issue was observed when there was no tolerations checkbox present on the dashboard. you can verify that in the images attached in this PR in original comment. @andrewballantyne

@tarukumar Ah I totally misunderstood the original issue and the bug I logged upstream.

Hmmm... there was a gateway issue we fixed in 1.16 -- perhaps you got a version before that fix was provided? It was fixed part of https://github.com/opendatahub-io/odh-dashboard/issues/354

🤔 I'm not quite sure how I misunderstood the issue looking back on the conversations now. heh

The distinction to know if you are before or after this fix is to check if you see the "Usage Data Collection" checkbox in a middle state `[-]` on load when you do not have any user tracking setting. I think in order to get into that state, you need to not have the ConfigMap `odh-segment-key-config` entirely in your Dashboard namespace. If it is there, the logic should resolve properly.

If it is unrelated, we may have another issue entirely.

tarukumar:
@anishasthana we observed that grafana is connected to Prometheus without tls.

Screenshot from 2022-08-26 15-27-23

pablofelix:
We also found this :
Screenshot from 2022-08-26 13-39-49

anishasthana:
Grafana is no longer required as part of the deployment,  so I wouldn't worry about that. Interesting that the secret wasn't created @atheo89 fyi

tarukumar:
> Grafana is no longer required as part of the deployment, so I wouldn't worry about that. Interesting that the secret wasn't created @atheo89 fyi

Just a thought shouldn't we remove it since it is no longer required?

anishasthana:
I agree. There is a jira open for it(trying to find it now), but it hasn't been targeted to any release and I'd rather not add additional changes to an already complex release.

atheo89:
This PR looks fine to me.

VaishnaviHire:
The code changes look good, will test them once the latest live-build is available

VaishnaviHire:
/lgtm 

1. Found another issue with Dashboard. Opened a Jira here : https://issues.redhat.com/browse/RHODS-4996
2. Monitoring issue mentioned [here](https://github.com/red-hat-data-services/odh-deployer/pull/256#discussion_r955190909) still persists .

VaishnaviHire (approved):

cfchase:
Crashloop backoff
[rhods-operator-69fb5548d6-j6d9q-rhods-deployer.log](https://github.com/red-hat-data-services/odh-deployer/files/9436626/rhods-operator-69fb5548d6-j6d9q-rhods-deployer.log)

deploy.sh
```diff
@@ -95,73 +101,81 @@ else
...
+
+  oc delete -n ${ODH_PROJECT} kfdef opendatahub --wait=false
+
+  ADMIN_GROUPS=$(oc get cm -n ${ODH_PROJECT} rhods-groups-config -o jsonpath="{.data.admin_groups}") || echo "rhods-groups-config not found"
+  ALLOWED_GROUPS=$(oc get cm -n ${ODH_PROJECT} rhods-groups-config -o jsonpath="{.data.allowed_groups}") || echo "rhods-groups-config not found"
+  CULLER_TIMEOUT=$(oc get cm -n ${ODH_PROJECT} jupyterhub-cfg -o jsonpath="{.data.culler_timeout}") || echo "jupyterhub-cfg not found"
+  DEFAULT_PVC_SIZE=$(oc get cm -n ${ODH_PROJECT} jupyterhub-cfg -o jsonpath="{.data.singleuser_pvc_size}") || echo "jupyterhub-cfgs not found"
+
+  if [ $OLD_CULLER_DEFAULT_TIMEOUT -ne $CULLER_TIMEOUT ]; then
+    sed -i "s/<culling_time>/$CULLER_TIMEOUT" nbc/notebook-controller-culler-config.yaml
```
cfchase: `sed -i "s/<culling_time>/$CULLER_TIMEOUT" nbc/notebook-controller-culler-config.yaml`

cfchase (approved):

anishasthana:
@pablofelix @tarukumar [quay.io/modh/rhods-operator-live-catalog:1.16.0-rhods-4760](http://quay.io/modh/rhods-operator-live-catalog:1.16.0-rhods-4760) is now the best live build to use.

tarukumar:
@anishasthana We performed migration from the above provided live build(quay.io/modh/rhods-operator-live-catalog:1.16.0-rhods-4760) and observed the below issue:

1. CRO Pod is not deleted after the migration(logs to be shared in the slack)
[takumar@fedora ~]$ oc get pod -n redhat-ods-operator
NAME                                      READY   STATUS    RESTARTS   AGE
cloud-resource-operator-5449867d4-w5dvh   1/1     Running   0          3h4m
rhods-operator-68ccf9bc4c-mbrdq           1/1     Running   0          51m

2. secrets related to CRO is not deleted
[takumar@fedora ~]$ oc get secrets -n redhat-ods-operator;oc get secrets -n redhat-ods-applications
NAME                                      TYPE                                  DATA   AGE
addon-managed-odh-parameters              Opaque                                1      3h5m
addon-managed-odh-pullsecret              kubernetes.io/dockerconfigjson        1      3h5m
builder-dockercfg-xgldj                   kubernetes.io/dockercfg               1      3h5m
builder-token-d99pn                       kubernetes.io/service-account-token   4      3h5m
builder-token-dlqvn                       kubernetes.io/service-account-token   4      3h5m
cloud-resource-operator-dockercfg-ggxn8   kubernetes.io/dockercfg               1      3h2m
cloud-resource-operator-token-5b7l8       kubernetes.io/service-account-token   4      3h2m
cloud-resource-operator-token-zcmhb       kubernetes.io/service-account-token   4      3h2m
default-dockercfg-pfdqq                   kubernetes.io/dockercfg               1      3h5m
default-token-7z2xj                       kubernetes.io/service-account-token   4      3h5m
default-token-qqql4                       kubernetes.io/service-account-token   4      3h5m
deployer-dockercfg-9bnlm                  kubernetes.io/dockercfg               1      3h5m
deployer-token-2xbqb                      kubernetes.io/service-account-token   4      3h5m
deployer-token-lftpn                      kubernetes.io/service-account-token   4      3h5m
rhods-operator-dockercfg-tvm7j            kubernetes.io/dockercfg               1      3h3m
rhods-operator-token-c62q9                kubernetes.io/service-account-token   4      3h3m
rhods-operator-token-ww27q                kubernetes.io/service-account-token   4      3h3m
NAME                                                  TYPE                                  DATA   AGE
addon-managed-odh-pullsecret                          kubernetes.io/dockerconfigjson        1      3h5m
anaconda-ce-access                                    Opaque                                0      167m
builder-dockercfg-h44qh                               kubernetes.io/dockercfg               1      3h5m
builder-token-7lfsf                                   kubernetes.io/service-account-token   4      3h5m
builder-token-88njt                                   kubernetes.io/service-account-token   4      3h5m
cloud-resources-aws-credentials                       Opaque                                3      3h2m
dashboard-oauth-config                                Opaque                                1      138m
dashboard-proxy-tls                                   kubernetes.io/tls                     2      138m
default-dockercfg-725rm                               kubernetes.io/dockercfg               1      3h5m
default-token-rm7c8                                   kubernetes.io/service-account-token   4      3h5m
default-token-z5snr                                   kubernetes.io/service-account-token   4      3h5m
deployer-dockercfg-7qqqf                              kubernetes.io/dockercfg               1      3h5m
deployer-token-g846g                                  kubernetes.io/service-account-token   4      3h5m
deployer-token-zzx5v                                  kubernetes.io/service-account-token   4      3h5m
notebook-controller-service-account-dockercfg-rhpzp   kubernetes.io/dockercfg               1      138m
notebook-controller-service-account-token-6z2ss       kubernetes.io/service-account-token   4      138m
notebook-controller-service-account-token-rm6rh       kubernetes.io/service-account-token   4      138m
odh-notebook-controller-manager-dockercfg-k9bzl       kubernetes.io/dockercfg               1      138m
odh-notebook-controller-manager-token-7kxcx           kubernetes.io/service-account-token   4      138m
odh-notebook-controller-manager-token-h49mq           kubernetes.io/service-account-token   4      138m
odh-notebook-controller-webhook-cert                  kubernetes.io/tls                     2      138m
odh-segment-key                                       Opaque                                1      167m
rhods-dashboard-dockercfg-x9ml4                       kubernetes.io/dockercfg               1      138m
rhods-dashboard-token-4wp6z                           kubernetes.io/service-account-token   4      138m
rhods-dashboard-token-9jb6h                           kubernetes.io/service-account-token   4      138m

anishasthana:
@tarukumar thanks for taking a look.  On Friday we elected to delete CRO completely in 1.17 this way the upgrade occurs a lot faster.

PR #327, opened 2023-04-14
Remove getting OVMS image from the CSV, use hardcoded

Xaenalt:
With the change to the quay.io image, we can the logic for getting this from the CSV

This is related to https://github.com/red-hat-data-services/odh-manifests/pull/328

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] For commits that came from upstream, `[UPSTREAM]` has been prepended to the commit message.
- [X] JIRA link(s): 
- [X] The Jira story is acked.
- [ ] Live build image: 
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -102,8 +102,8 @@ function oc::object::safe::to::apply() {
 }
 
 function add_servingruntime_config() {
-  # Get OpenVino image from latest CSV
-  openvino_img=$(oc get csv -l operators.coreos.com/rhods-operator.redhat-ods-operator -n $ODH_OPERATOR_PROJECT -o=jsonpath='{.items[-1].spec.install.spec.deployments[?(@.name == "rhods-operator")].spec.template.spec.containers[?(@.name == "rhods-operator")].env[?(@.name == "RELATED_IMAGE_ODH_OPENVINO_IMAGE")].value}')
+  # Replace OpenVINO Model Server image in servingruntime config
+  openvino_img='quay.io/modh/openvino-model-server@sha256:c89f76386bc8b59f0748cf173868e5beef21ac7d2f78dada69089c4d37c44116'
     
   # Replace image
   sed -i "s|<openvino_image>|${openvino_img}|g" model-mesh/serving_runtime_config.yaml
--- a/kfdefs/rhods-model-mesh.yaml
+++ b/kfdefs/rhods-model-mesh.yaml
@@ -12,8 +12,6 @@ spec:
           value: ${RELATED_IMAGE_ODH_MODELMESH_RUNTIME_ADAPTER_IMAGE}
         - name: odh-modelmesh
           value: ${RELATED_IMAGE_ODH_MODELMESH_IMAGE}
-        - name: odh-openvino
-          value: ${RELATED_IMAGE_ODH_OPENVINO_IMAGE}
         - name: odh-modelmesh-controller
           value: ${RELATED_IMAGE_ODH_MODELMESH_CONTROLLER_IMAGE}
         - name: odh-model-controller


lucferbux:
We still ned this right? I assume sha of that related image could change between versions, hence we need to update it when upgrading the operator right?

And this logic will change in model serving v2.

zdtsw (approved):
/hold

deploy.sh
```diff
@@ -102,8 +102,8 @@ function oc::object::safe::to::apply() {
 }
 
 function add_servingruntime_config() {
-  # Get OpenVino image from latest CSV
-  openvino_img=$(oc get csv -l operators.coreos.com/rhods-operator.redhat-ods-operator -n $ODH_OPERATOR_PROJECT -o=jsonpath='{.items[-1].spec.install.spec.deployments[?(@.name == "rhods-operator")].spec.template.spec.containers[?(@.name == "rhods-operator")].env[?(@.name == "RELATED_IMAGE_ODH_OPENVINO_IMAGE")].value}')
+  # Replace OpenVINO Model Server image in servingruntime config
+  openvino_img='quay.io/modh/openvino-model-server@sha256:c89f76386bc8b59f0748cf173868e5beef21ac7d2f78dada69089c4d37c44116'
```
lucferbux, line 106: I'm sorry, I still don't understand this change, why would we want to harcode the value here? Is it not present in the csv too? https://gitlab.cee.redhat.com/data-hub/rhods-cpaas-midstream/-/blob/rhods-1.26-rhel-8/distgit/containers/odh-operator-bundle/bundle-template/manifests/rhods-operator.clusterserviceversion.yml#L667

that would mean that we should update it every time there's a rebuild, who would be in charge of that?

I'm still not sure why we want to change this logic.

lucferbux (changes requested):

deploy.sh
lucferbux: @zdtsw you might be able to answer this question, I'm sorry I don't have the hole context here

kfdefs/rhods-model-mesh.yaml
```diff
@@ -12,8 +12,6 @@ spec:
           value: ${RELATED_IMAGE_ODH_MODELMESH_RUNTIME_ADAPTER_IMAGE}
         - name: odh-modelmesh
           value: ${RELATED_IMAGE_ODH_MODELMESH_IMAGE}
-        - name: odh-openvino
```
lucferbux, line 15: And why are we removing this? Isn't openivno a related image still?

lucferbux (changes requested):

deploy.sh
zdtsw: right.
so my understanding:

1) openvino image is unlikely to change in a frequent fashion. (i.e maybe several months we will build a new openvino image by openshift-ci) so if we hard code digest , it is not that often we need to update it. 

2) the reason why hard code it and set to `openvino_img=` is to not set the RELATED_IMAGE_* variable in CSV file.
and only add openvino pullspec (hardcode one) in the CSV ` spec.annotation`s and let the build system to auto generate and inject it into `relatedImages` 

But after last Friday, some internal discussion of how our build system works.
I think the solution to get it into annotations wont work as well, due to the fact where openvino image lives. 

so we need to hold this PR for a while, till we find a way that works with build and follow that direction to change our solution in deploy and CSV and so on.

zdtsw (commented):
lets wait a bit till more clear how it can work with our build

kfdefs/rhods-model-mesh.yaml
Xaenalt: No, the goal is to move this out of related images, since CPaaS won't let us use a quay image in that field

deploy.sh
Xaenalt: Correct, the reason we're moving it is that CPaaS has a check in the relatedImages field to prevent non-approved image repos, which quay is not one of those. They're reluctant to add quay because of the implications system-wide, so this ends up being the only way

Xaenalt:
@LaVLaS @VaishnaviHire PTAL

VaishnaviHire:
/lgtm 

Given the entry under the [CSV](https://gitlab.cee.redhat.com/data-hub/rhods-cpaas-midstream/-/blob/rhods-1.26-rhel-8/distgit/containers/odh-operator-bundle/bundle-template/manifests/rhods-operator.clusterserviceversion.yml#L667) is removed

VaishnaviHire (approved):

deploy.sh
Jooho: The best way is... put this value in one place and use it everywhere like before. Sha in CSV and other components use it from CSV. But CPaaS does not like quay.io so we need to find another place but I couldn't find a good place. So for now, hard coded would be good to go but we need to revisit this.
For your information, this is the reason why we can not put NOTEBOOK sha into CSV. please refer this for notebook @VaishnaviHire  Last time, I heard you have the plan to put NOTEBOOK sha into environmental variables in the CSV. But CPaaS does not allow it.

kfdefs/rhods-model-mesh.yaml
Jooho: CPaaS does not allow to add quay.io image into it. We need to update disconnected installation official doc as well.

Jooho (approved):
/lgtm

Jooho:
@Xaenalt It needs to update a doc for disconnected installation. Could you please check this as well?

Xaenalt:
@Jooho Yeah, I'll add a ping in the JIRA about it to the docs team

deploy.sh
lucferbux: Ok, great, bear in mind this will change with model serving v2, I suggest we should add this into the default ovms runtime object once we have it, thanks all for the update, I didn't have the context.

lucferbux (approved):

PR #334, opened 2023-05-02
Add an alert for detecting stuck deployment rollouts

VannTen:
## Description

Alerts of deployment not progressing (whatever the reason) for ODH controllers.

## Merge criteria:

- [x] JIRA link(s): https://issues.redhat.com/browse/RHODS-8152
- [x] Live build image: quay.io/modh/rhods-operator-live-catalog:1.27.0-rhods-8152

### DIFF
--- a/monitoring/prometheus/prometheus-configs.yaml
+++ b/monitoring/prometheus/prometheus-configs.yaml
@@ -706,6 +706,14 @@ data:
           for: 5m
           labels:
             severity: warning
+        - alert: Controller rollout is stuck
+          annotations:
+            message: 'Rollout of controller {{ $labels.deployment }} is stuck.'
+            summary: A RHODS controller deployment upgrade is stuck.
+          expr: sum(kube_deployment_status_condition{condition="Progressing", status="false"}) by (deployment) != 0
+          for: 30m
+          labels:
+            severity: warning
 
   prometheus.yml: |
     rule_files:


monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -706,6 +706,14 @@ data:
           for: 5m
           labels:
             severity: warning
+        - alert: Controller rollout is stuck
+          annotations:
+            message: 'Rollout of controller {{ $labels.deployment }} is stuck.'
+            summary: An ODH controller deployment upgrade is stuck.
+          expr: sum(kube_deployment_status_condition{condition="Progressing", status="false"}) by (deployment) != 0
```
anishasthana, line 713: Won't we need to filter these to only alert for RHODS pods?

monitoring/prometheus/prometheus-configs.yaml
```diff
@@ -706,6 +706,14 @@ data:
           for: 5m
           labels:
             severity: warning
+        - alert: Controller rollout is stuck
+          annotations:
+            message: 'Rollout of controller {{ $labels.deployment }} is stuck.'
+            summary: An ODH controller deployment upgrade is stuck.
```
anishasthana, line 712: Should say RHODS :-)

anishasthana (changes requested):

monitoring/prometheus/prometheus-configs.yaml
VannTen: Maybe, I have a bit of doubt about that:
- This rules are against the prometheus instance shipped by RHODS, right ? -> in that case I'd rather filter directly in the collect job which takes those metrics from the cluster-prometheus.
- If not, we can just filter by namespace: `{namespace~='redhat-ods-applications|<etc>}`

VannTen:
See kubernetes-monitoring/kubernetes-mixin#845 for a potential more general alternative.

VannTen:
/close
(as the upstream PR was merged, it will flow down into openshift eventually)

PR #340, opened 2023-05-19
Update servingruntime ootb configuration

lucferbux:
## Description

Get latest changes to downstream

### DIFF
--- a/odh-dashboard/modelserving/ovms-gpu-ootb.yaml
+++ b/odh-dashboard/modelserving/ovms-gpu-ootb.yaml
@@ -1,0 +1,60 @@
+kind: Template
+apiVersion: template.openshift.io/v1
+metadata:
+  name: ovms-gpu
+  labels:
+    opendatahub.io/dashboard: 'true'
+    opendatahub.io/ootb: 'true'
+    opendatahub.io/configurable: 'true'
+  annotations:
+    tags: 'ovms,servingruntime'
+    description: 'OpenVino with GPU Support Model Serving Definition'
+objects:
+  - apiVersion: serving.kserve.io/v1alpha1
+    kind: ServingRuntime
+    metadata:
+      name: ovms-gpu
+      annotations:
+        openshift.io/display-name: 'OpenVINO Model Server (Supports GPUs)'
+      labels:
+        opendatahub.io/dashboard: 'true'
+    spec:
+      builtInAdapter:
+        env:
+          - name: OVMS_FORCE_TARGET_DEVICE
+            value: NVIDIA
+        memBufferBytes: 134217728
+        modelLoadingTimeoutMillis: 90000
+        runtimeManagementPort: 8888
+        serverType: ovms
+      containers:
+        - args:
+            - '--port=8001'
+            - '--rest_port=8888'
+            - '--config_path=/models/model_config_list.json'
+            - '--file_system_poll_wait_seconds=0'
+            - '--grpc_bind_address=127.0.0.1'
+            - '--rest_bind_address=127.0.0.1'
+          image: ovms-1
+          name: ovms
+          resources:
+            limits:
+              cpu: '0'
+              memory: 0Gi
+            requests:
+              cpu: '0'
+              memory: 0Gi
+      grpcDataEndpoint: 'port:8001'
+      grpcEndpoint: 'port:8085'
+      multiModel: true
+      protocolVersions:
+        - grpc-v1
+      replicas: 1
+      supportedModelFormats:
+        - autoSelect: true
+          name: openvino_ir
+          version: opset1
+        - autoSelect: true
+          name: onnx
+          version: '1'
+parameters: []
--- a/odh-dashboard/modelserving/ovms-gpu-ootb.yaml
+++ b/odh-dashboard/modelserving/ovms-gpu-ootb.yaml
@@ -57,4 +57,7 @@ objects:
         - autoSelect: true
           name: onnx
           version: '1'
+        - autoSelect: true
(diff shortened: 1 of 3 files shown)


lucferbux:
@Xaenalt @VaishnaviHire @andrewballantyne 
Can we review this before feature freeze please, in that case this can go to 1.28 without opening another pr.
thanks in advance!

odh-dashboard/modelserving/ovms-ootb.yaml
```diff
@@ -16,6 +16,7 @@ objects:
       name: ovms
       annotations:
         openshift.io/display-name: 'OpenVINO Model Server'
+        opendatahub.io/disable-gpu: 'true'
```
andrewballantyne, line 19: FWIW, for Habana integration we are talking about this being more robust. Something like `opendatahub.io/accelerator-support` and it takes a JSON of the string that we would put into the container resources.

No need to change that now, just an FYI for where we are going.

odh-dashboard/modelserving/ovms-gpu-ootb.yaml
```diff
@@ -0,0 +1,60 @@
...
+apiVersion: template.openshift.io/v1
+metadata:
+  name: ovms-gpu
+  labels:
+    opendatahub.io/dashboard: 'true'
+    opendatahub.io/ootb: 'true'
+    opendatahub.io/configurable: 'true'
+  annotations:
+    tags: 'ovms,servingruntime'
+    description: 'OpenVino with GPU Support Model Serving Definition'
```
andrewballantyne, line 11: Interesting this is not `openshift.io/description` 🤔

@lucferbux do you clone these items from the Template over when they clone it in the UI?

odh-dashboard/modelserving/ovms-ootb.yaml
lucferbux: Thanks!

Yes, this is a workaround, I hope that once we implement havana, we'll have a more robust gpu integration for Openvino

odh-dashboard/modelserving/ovms-gpu-ootb.yaml
lucferbux: No! That's a great feedback, we were using this in the ootb templates since it's one of the attributes in the `Template` object.
We are not using descriptions anywhere in model serving but we can either remove this description or add the logic to add it to the ServingRuntime

andrewballantyne: It's an annotation -- `description` may or may not be something Templates are known for having -- but I don't think it's part of the Template in any official way.

I caution using anything on the Template in the cloned Template because we just don't show these values anywhere -- it'll be hard for the Admin to modify them.

Xaenalt:
Oh, this seems like a good place to ask, apparently despite the docs saying this isn't supported, apparently we can set target_device: HETERO:NVIDIA,CPU, so it'll fall back if it fails to compile onto the GPU, I'm not entirely sure if this will work if the Nvidia device is absent, but it's worth asking if we want to look into that

lucferbux:
> Oh, this seems like a good place to ask, apparently despite the docs saying this isn't supported, apparently we can set target_device: HETERO:NVIDIA,CPU, so it'll fall back if it fails to compile onto the GPU, I'm not entirely sure if this will work if the Nvidia device is absent, but it's worth asking if we want to look into that

Hi @Xaenalt I don't fully understand this, what do you mean? In case they deploy this serving runtime and nvidia is absent it will fail? Is that it?

Xaenalt:
So, I need to test to make sure what happens if Nvidia is absent, but if I'm reading their docs correctly (and what their devs said correctly), it'll run some parts on the CPU and some on the Nvidia GPU, that should increase compatibility, but it's not clear to me what'll happen if one device is absent, I'd assume that it'd just run it all on the CPU, but I've been burned by that kind of assumption before xD

lucferbux:
> So, I need to test to make sure what happens if Nvidia is absent, but if I'm reading their docs correctly (and what their devs said correctly), it'll run some parts on the CPU and some on the Nvidia GPU, that should increase compatibility, but it's not clear to me what'll happen if one device is absent, I'd assume that it'd just run it all on the CPU, but I've been burned by that kind of assumption before xD

But does this require something UI related? or maybe adding something in the official docs? Not sure if I need to change something based on your comments.

lucferbux:
@Xaenalt added https://github.com/red-hat-data-services/odh-deployer/pull/340/commits/fe33221e4e034359469c211c92355d1f035922e8 to support tensorflow, let me know if this is fine!

andrewballantyne (approved):
From my perspective -- this lgtm.

odh-dashboard/modelserving/ovms-gpu-ootb.yaml
```diff
@@ -0,0 +1,63 @@
...
+      supportedModelFormats:
+        - autoSelect: true
+          name: openvino_ir
+          version: opset1
+        - autoSelect: true
+          name: onnx
+          version: '1'
+        - name: tensorflow
+          version: "2"
+          autoSelect: true
```
Xaenalt, line 62: Nit, this is a different order than the order of the other two, autoSelect is first in the others

lucferbux:
@anishasthana @VaishnaviHire @LaVLaS could you check this out? I'm aiming to merge this before feature freeze, thanks in advance!

Xaenalt (approved):

lucferbux:
@tarukumar I think it needs qe approval too, can you please take a look?

PR #358, opened 2023-06-09
Fix isv rearrange order

lucferbux:
## Description

Follow up of https://issues.redhat.com/browse/RHODS-9226 introduced a bug where some of the shared tiles were misconfigured

## How Has This Been Tested?

1. Install the operator on a managed cloud
2. Check that starbust enterprise is not installed
3. Check that the rest of applications are there

Screenshot 2023-06-09 at 09 30 34

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] For commits that came from upstream, `[UPSTREAM]` has been prepended to the commit message.
- [X] JIRA link(s): https://issues.redhat.com/browse/RHODS-9226
- [X] The Jira story is acked.
- [X] Live build image: `quay.io/lferrnan/rhods-operator-live-catalog:1.28.0-deployer`
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work.

PR #359, opened 2023-06-09
Fix isv rearrange order

lucferbux:
Follow up of https://github.com/red-hat-data-services/odh-deployer/pull/358 for 1.28


zdtsw (approved):
/lgtm

zdtsw:
we've checked with QE with confirmation the new lists of all ISV applied to different installation env. is correct.

PR #360, opened 2023-06-12
Fix startbust enablement when installing operator

lucferbux:
## Description

Fix issue when installing starbust enterprise the app didn't appear as enabled.

## How Has This Been Tested?

1. Install startbus enterprise operator
2. Check that the app has been enabled

Screenshot 2023-06-13 at 17 32 05

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] JIRA link(s): https://issues.redhat.com/browse/RHODS-9327
- [X] The Jira story is acked.
- [X] Live build image: `quay.io/lferrnan/rhods-operator-live-catalog:1.28.0-deployer`
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work.

### DIFF
--- a/odh-dashboard/apps-on-prem/starburst-enterprise/starburstenterprise-app.yaml
+++ b/odh-dashboard/apps-on-prem/starburst-enterprise/starburstenterprise-app.yaml
@@ -15,6 +15,7 @@ spec:
   route:
   category: Self-managed
   support: Starburst
+  csvName: starburst-enterprise-helm-operator
   docsLink: https://docs.starburst.io/latest/overview.html
   quickStart: using-starburst-enterprise
   getStartedLink: https://docs.starburst.io/ecosystems/redhat/openshift-deployment.html


lucferbux:
@bdattoma It would be nice to confirm this with Karl, we don't need to add this logic to starburst for managed right?

bdattoma:
> managed

when you say "managed", are you referring to Managed Starburst or Managed RHODS? if you refer to Managed RHODS, I think it is not needed because Starburst Galaxy cannot be enabled in the cluster, it's an external service. What we could do is to add the "Enable" button like it was done for RHOSAK. 

if you were referring to Managed Starburst, then yes we will need to add the same logic, but for now there is no tile for it in RHODS.

PM should give a final call on both the points.

lucferbux:
> > managed
> 
> when you say "managed", are you referring to Managed Starburst or Managed RHODS? if you refer to Managed RHODS, I think it is not needed because Starburst Galaxy cannot be enabled in the cluster, it's an external service. What we could do is to add the "Enable" button like it was done for RHOSAK, which would not be a real installation, just a way to make it appearing in "Enabled" page of RHODS.
> 
> if you were referring to Managed Starburst, then yes we will need to add the same logic, but for now there is no tile for it in RHODS.
> 
> PM should give a final call on both the points.

Yeah, I'm sorry, terminology is quite tricky, I was talking about Manage RHODS, meaning the Starbust Galaxy deployment. I was specifically asking if we should add something to that `app` resource.

But we might wanna add the `enable` mechanism that you refer. we can hop back to the slack conversation just to confirm this.

zdtsw (approved):
/lgtm

